PR #2, opened 2025-03-15
General Updates

icywei:
GNB - PVE Updates
Added Burst Window Advance Option

PLD - PVE Updates
Update rotations
Prevent Clemency while moving
Added Burst Window Advance Option

VPR - PVE Updates
Update rotations
Added Burst Window Advance Option


icywei:
Got bugs

PR #77, opened 2025-08-24
Localize more strings

mastahg:


cheesegoldfish:
lmk when this is ready to merge.

mastahg:
This should be ready. There's still more work to be done for headers but this should be all the main content.

PR #196, opened 2026-07-30
Magitek - Revive fallen party members with Phoenix Down

HomoYouDidnt:
Uses Phoenix Downs to revive fallen party members and Duty Support NPCs in instanced content, as a last resort after the job's own raise.

- Runs only in instanced content, deferring to the game's own item gating for which duties permit it — the assumption that the item's usability check fully encodes duty legality has not been separately verified in every duty type
- Keeps a configurable reserve so the routine never spends a player's last few
- Waits out the death delay first, so a real raise always takes priority
- Orders targets healers, then tanks, then DPS
- Waits for the item cast to finish rather than treating the start as done, which previously let the next pulse walk away and cancel the revive
- Skips corpses whose object has been freed, which could otherwise throw out of the whole heal step

Off by default. Test plan: enable in a Duty Support dungeon with raises disabled, let an NPC or party member die, and confirm one Phoenix Down is used after the configured delay, the reserve floor is respected, and no second down is spent on the same corpse while the revive registers.

### DIFF
--- a/Magitek/Logic/Roles/PhoenixDown.cs
+++ b/Magitek/Logic/Roles/PhoenixDown.cs
@@ -22,8 +22,9 @@ namespace Magitek.Logic.Roles
 
         // Per-target throttle. After a Phoenix Down is used the revive takes a server round-trip to land,
         // during which the target still reads CurrentHealth==0 — skip it briefly so we don't fire a
-        // second one before the first registers. Kept well under the death delay so it can never block a
-        // legitimate re-revive (that needs a fresh death plus the full delay again).
+        // second one before the first registers. With the default death delay a legitimate re-revive
+        // needs a fresh death plus the full delay again, which outlasts this window; at very low
+        // user-set delays a rapid re-death may wait out the remainder of the 10s before a second down.
         private static readonly Dictionary<uint, DateTime> _recentUses = new();
         private static readonly TimeSpan UseThrottle = TimeSpan.FromSeconds(10);
 
--- a/Magitek/Logic/Roles/PhoenixDown.cs
+++ b/Magitek/Logic/Roles/PhoenixDown.cs
@@ -20,14 +20,6 @@ namespace Magitek.Logic.Roles
         private const uint PhoenixDownItemId = 4570;
         private const float PhoenixDownRange = 15f; // Phoenix Down reaches ~15 yalms.
 
-        // Per-target throttle. After a Phoenix Down is used the revive takes a server round-trip to land,
-        // during which the target still reads CurrentHealth==0 — skip it briefly so we don't fire a
-        // second one before the first registers. With the default death delay a legitimate re-revive
-        // needs a fresh death plus the full delay again, which outlasts this window; at very low
-        // user-set delays a rapid re-death may wait out the remainder of the 10s before a second down.
-        private static readonly Dictionary<uint, DateTime> _recentUses = new();
-        private static readonly TimeSpan UseThrottle = TimeSpan.FromSeconds(10);
-
         public static async Task<bool> Execute()
         {
             if (!BaseSettings.Instance.UsePhoenixDown)
@@ -55,8 +47,8 @@ namespace Magitek.Logic.Roles
             // is built from PartyManager.RawMembers). Skip anyone already being raised, require the death
             // to be at least `delaySeconds` old so a real raise gets priority, and require the game to
             // actually allow the use on them (item.CanUse -> real range / line-of-sight / duty rules) so
-            // an unusable higher-priority body can't starve a reachable one. Honour the per-target
-            // throttle. GetResurrectionWeight orders healers first, then tanks, then DPS.
+            // an unusable higher-priority body can't starve a reachable one. GetResurrectionWeight
+            // orders healers first, then tanks, then DPS.
             // u.IsValid short-circuits before every read below. Each of those touches game memory, so a
             // corpse whose object was freed mid-iteration throws ReadWriteMemoryException from inside the
             // predicate and takes down the whole Heal pulse — the same guard Healer.ResurrectionLogic
@@ -70,7 +62,6 @@ namespace Magitek.Logic.Roles
                             && u.IsTargetable
                             && u.WithinSpellRange(PhoenixDownRange)
                             && item.CanUse(u)
-                            && (!_recentUses.TryGetValue(u.ObjectId, out var last) || now - last > UseThrottle)
                             && Group.GetDeathTime(u)?.AddSeconds(delaySeconds) <= now)
                 .OrderByDescending(u => u.GetResurrectionWeight())
                 .FirstOrDefault();
@@ -80,30 +71,21 @@ namespace Magitek.Logic.Roles
 
             // One targeted use per pulse. NOT the shared UseItem() extension: it loops UseItem() with no
             // target and no inter-use delay, which would fire on the wrong unit and re-fire every frame
-            // until the revive lands (burning multiple Phoenix Downs). Stamp the throttle and use once;
-            // the frame-pulsed Execute re-checks the reserve/target next tick.
-            _recentUses[target.ObjectId] = now;
-            PruneThrottle(now);
-
+            // until the revive lands (burning multiple Phoenix Downs).
             Logger.WriteInfo($"[Phoenix Down] Reviving {target.Name} ({target.CurrentJob})");
             item.UseItem(target);
 
             // Starting the item cast is not the same as finishing it. Left untracked, the very next pulse
             // sees nothing in progress and carries on to the movement step, which walks and cancels the
-            // revive. Hold here until the cast actually ends so the rest of the pulse cannot interrupt it.
+            // revive. Hold here until the cast actually ends so the rest of the pulse cannot interrupt
+            // it — this hold, plus the !HasAura(Raise) filter above, is the whole re-fire guard: no
+            // per-corpse timers. The residual window (cast end to the server's Raise status) is one
+            // round-trip; a second down inside it is possible but rare, the same trade the non-party
+            // rez guard makes.
             await Coroutine.Wait(1000, () => Core.Me.IsCasting);
             await Coroutine.Wait(10000, () => !Core.Me.IsCasting);
 
             return true;
-        }
-
-        private static void PruneThrottle(DateTime now)
-        {
-            if (_recentUses.Count < 16)
-                return;
-
-            foreach (var stale in _recentUses.Where(kv => now - kv.Value > UseThrottle).Select(kv => kv.Key).ToList())
-                _recentUses.Remove(stale);
         }
     }
 }
--- a/Magitek/Logic/Roles/PhoenixDown.cs
+++ b/Magitek/Logic/Roles/PhoenixDown.cs
@@ -56,8 +56,14 @@ namespace Magitek.Logic.Roles
             // actually allow the use on them (item.CanUse -> real range / line-of-sight / duty rules) so
             // an unusable higher-priority body can't starve a reachable one. Honour the per-target
             // throttle. GetResurrectionWeight orders healers first, then tanks, then DPS.
+            // u.IsValid short-circuits before every read below. Each of those touches game memory, so a
+            // corpse whose object was freed mid-iteration throws ReadWriteMemoryException from inside the
+            // predicate and takes down the whole Heal pulse — the same guard Healer.ResurrectionLogic
+            // already carries for this list.
             var target = Group.DeadAllies
-                .Where(u => u.CurrentHealth == 0
+                .Where(u => u != null
+                            && u.IsValid
+                            && u.CurrentHealth == 0
                             && !u.HasAura(Auras.Raise)
                             && u.IsVisible
                             && u.IsTargetable
--- a/Magitek/Logic/Roles/PhoenixDown.cs
+++ b/Magitek/Logic/Roles/PhoenixDown.cs
@@ -1,3 +1,4 @@
+using Buddy.Coroutines;
 using ff14bot;
 using ff14bot.Managers;
 using Magitek.Extensions;
@@ -79,6 +80,13 @@ namespace Magitek.Logic.Roles
 
             Logger.WriteInfo($"[Phoenix Down] Reviving {target.Name} ({target.CurrentJob})");
             item.UseItem(target);
+
+            // Starting the item cast is not the same as finishing it. Left untracked, the very next pulse
+            // sees nothing in progress and carries on to the movement step, which walks and cancels the
+            // revive. Hold here until the cast actually ends so the rest of the pulse cannot interrupt it.
+            await Coroutine.Wait(1000, () => Core.Me.IsCasting);
+            await Coroutine.Wait(10000, () => !Core.Me.IsCasting);
+
             return true;
         }
 
--- a/Magitek/Logic/Roles/PhoenixDown.cs
+++ b/Magitek/Logic/Roles/PhoenixDown.cs
@@ -5,7 +5,6 @@ using Magitek.Extensions;
 using Magitek.Models.Account;
 using Magitek.Utilities;
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Auras = Magitek.Utilities.Auras;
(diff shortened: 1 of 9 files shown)


Magitek/Logic/Roles/PhoenixDown.cs
```diff
@@ -0,0 +1,94 @@
...
+            // One targeted use per pulse. NOT the shared UseItem() extension: it loops UseItem() with no
+            // target and no inter-use delay, which would fire on the wrong unit and re-fire every frame
+            // until the revive lands (burning multiple Phoenix Downs). Stamp the throttle and use once;
+            // the frame-pulsed Execute re-checks the reserve/target next tick.
+            _recentUses[target.ObjectId] = now;
+            PruneThrottle(now);
+
+            Logger.WriteInfo($"[Phoenix Down] Reviving {target.Name} ({target.CurrentJob})");
+            item.UseItem(target);
+            return true;
```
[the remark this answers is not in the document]
HomoYouDidnt: Fixed. It now waits for the item cast to finish rather than treating starting it as done.

Magitek/Models/Account/BaseSettings.cs
```diff
@@ -582,5 +582,17 @@ public static void ValidateSettingsWindowPosition(double windowWidth = 1000, dou
...
+        [DefaultValue(false)]
+        public bool UsePhoenixDown { get; set; }
+
+        [Setting]
+        [DefaultValue(30)]
+        public int PhoenixDownDelaySeconds { get; set; }
+
+        [Setting]
+        [DefaultValue(3)]
+        public int PhoenixDownReserve { get; set; }
```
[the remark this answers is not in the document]
HomoYouDidnt: Checked this one and the premise does not hold, so I have left the default alone.

Phoenix Down (item 4570) is not Unique. Game data gives `IsUnique: false` and `StackSize: 999` — it is ordinary stackable Medicine, so a character can carry up to 999 and `owned` is not capped at 1.

It also demonstrably works as shipped. A live session on this build, with the reserve at its default of 3, reviving twice:

```
[04:49:28] [Magitek] [Phoenix Down] Reviving Alphinaud (Sage)
[05:18:27] [Magitek] [Phoenix Down] Reviving G'raha Tia (BlackMage)
```

The reserve is doing what it is meant to: hold back the last few so the routine does not spend a player's final Phoenix Downs without being asked. Anyone who wants every last one used can set it to 0, which the setting explicitly supports.

Worth noting the same session did surface a real defect in this file, now fixed here: the dead-ally search read eight properties off each corpse with no validity guard, so a Duty Support NPC freed mid-enumeration threw ReadWriteMemoryException out of the predicate and killed the whole Heal pulse. Healer.cs already carried that guard on the same list; this did not.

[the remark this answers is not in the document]
HomoYouDidnt: Correcting my own evidence above. The two log lines I cited do not show what I said they showed.

They are `[Phoenix Down] Reviving ...` entries, which Magitek writes when it *selects* a target and calls `UseItem` — not when a revive lands. Checking the combat log at both timestamps, the character was in a public zone surrounded by unrelated players (Peloton, En Avant, Sprint from three different strangers), not in a duty with Duty Support. The same log has no instance entry anywhere near either timestamp.

So those were casts at stale `Group.DeadAllies` entries, not confirmed revives. Citing them as proof the feature "demonstrably works as shipped" was wrong and I should not have made that claim without checking the other log.

What that does not change is the reason for closing this: Phoenix Down (item 4570) is `IsUnique: false` with `StackSize: 999`, so `owned` is not capped at 1 and the reserve-of-3 premise does not hold. That part stands on game data alone and is why the default is unchanged.

What it does change is my confidence that the revive path has ever been observed working end to end. It has not, in the data I have. It also fits the crash fixed in this PR rather too neatly — the routine was picking targets out of a list holding references that no longer matched anything live, and two minutes after the second of those two attempts one of them was freed mid-enumeration and threw ReadWriteMemoryException out of the predicate.

Magitek/Logic/Roles/PhoenixDown.cs
```diff
@@ -0,0 +1,91 @@
...
+            if (!BaseSettings.Instance.UsePhoenixDown)
+                return false;
+
+            // Instanced content only (dungeons/trials/deep dungeons/field ops/etc.). The game's own
+            // item.CanUse() then enforces which specific duties actually permit Phoenix Down and when
+            // (e.g. blocked in combat in 8-player content), so we don't hardcode those rules.
+            if (!DutyManager.InInstance)
+                return false;
+
+            var item = InventoryManager.FilledSlots.FirstOrDefault(r => r.RawItemId == PhoenixDownItemId);
```
cheesegoldfish: Ordering rather than correctness, but it's the hot path. These are the two most expensive calls in the method and they run before anything cheap can bail out. Measured on a live client with 138 filled slots:

```
FilledSlots.Where(...).Sum(...)                ~317 us
FilledSlots.FirstOrDefault(r => RawItemId...)  ~290 us
FilledSlots.Count()   (walk only)              ~202 us
DutyManager.InInstance                           0.022 us
Group.DeadAllies.Count                           0.012 us
```

`FilledSlots` walks every slot with a memory read each — which is why `FirstOrDefault` costs *more* than `Count()`. So `Heal()` pays ~0.6 ms per pulse scanning the bags twice, and in the overwhelming majority of pulses nobody is dead and the result was never going to be used.

`Group.DeadAllies` is a plain `List<Character>` already built by `Group.UpdateAllies()` earlier in the same pulse, so `.Count` is O(1) — roughly 25,000x cheaper than one bag scan. Gating on it first, and collapsing the two scans into one:

```csharp
if (!BaseSettings.Instance.UsePhoenixDown) return false;
if (Group.DeadAllies.Count == 0) return false;   // O(1); short-circuits almost every pulse
if (!DutyManager.InInstance) return false;

var slots = InventoryManager.FilledSlots.Where(r => r.RawItemId == PhoenixDownItemId).ToList();
if (slots.Count == 0) return false;
if (slots.Sum(r => (long)r.Count) <= BaseSettings.Instance.PhoenixDownReserve) return false;
var item = slots[0];
```

The cheap-first trick only works on `.Count` — running the full `DeadAllies` predicate below first would be worse, since that reads eight properties per corpse.

Magitek/Logic/Roles/PhoenixDown.cs
```diff
@@ -0,0 +1,91 @@
...
+            // predicate and takes down the whole Heal pulse — the same guard Healer.ResurrectionLogic
+            // already carries for this list.
+            var target = Group.DeadAllies
+                .Where(u => u != null
+                            && u.IsValid
+                            && u.CurrentHealth == 0
+                            && !u.HasAura(Auras.Raise)
+                            && u.IsVisible
+                            && u.IsTargetable
+                            && u.WithinSpellRange(PhoenixDownRange)
```
cheesegoldfish: `WithinSpellRange` is the one case AGENTS.md carves out an exception for, and revive targets are exactly it. `Logic/Roles/Healer.cs` carries the reasoning, on this same `Group.DeadAllies` list:

> Raise range is a raw 3D distance on purpose... A corpse has no usable CombatReach for the game to measure against, so raise range is centre to centre. WithinSpellRange would subtract both hitboxes and, because it uses Distance2D, ignore height entirely — calling a body far below us in range. Leave this as Distance.

Both of those errors run *permissive*, so this admits corpses that are genuinely out of range and leaves `item.CanUse` as the only thing stopping us acting on them. At 15 yalms the height blindness matters more than it does at raise's 30, not less.

```csharp
&& u.Distance(Core.Me) <= PhoenixDownRange
```

Related: `Healer.cs` also filters `u.InLineOfSight()` on this list, and that's missing here. If `item.CanUse` genuinely covers line of sight it's worth saying so with a reference — the comment above asserts it but nothing verifies it.

Magitek/Logic/Roles/PhoenixDown.cs
```diff
@@ -0,0 +1,91 @@
...
+            item.UseItem(target);
+
+            // Starting the item cast is not the same as finishing it. Left untracked, the very next pulse
+            // sees nothing in progress and carries on to the movement step, which walks and cancels the
+            // revive. Hold here until the cast actually ends so the rest of the pulse cannot interrupt
+            // it — this hold, plus the !HasAura(Raise) filter above, is the whole re-fire guard: no
+            // per-corpse timers. The residual window (cast end to the server's Raise status) is one
+            // round-trip; a second down inside it is possible but rare, the same trade the non-party
+            // rez guard makes.
+            await Coroutine.Wait(1000, () => Core.Me.IsCasting);
```
cheesegoldfish: This is the change I'd most like. Codex was right that starting the cast isn't finishing it, but holding the coroutine treats the symptom.

Phoenix Down has an **8 second cast** ([wiki](https://ffxiv.consolegameswiki.com/wiki/Phoenix_Down)), so this blocks the entire Heal pulse for 8 seconds — no healing on the living, no fight logic, no defensives, no tankbuster response — while we revive one person. On a dungeon pull that's likely to cost someone else.

Magitek already has non-blocking machinery for this. `Casting.TrackSpellCast()` runs at the top of `Heal()` (`RotationManager.cs:222`) and holds the pulse for an in-flight cast — but it explicitly disowns anything Magitek didn't record:

```csharp
if (!CastingTime.IsRunning) return false;   // Casting.cs
```

`item.UseItem()` sets none of that, which is exactly why the next pulse walked to movement. Recording the cast the way `SpellDataExtensions.Cast()` does (`SpellDataExtensions.cs:69-73`) lets the existing guard hold it, without blocking:

```csharp
if (!item.UseItem(target))
    return false;

Casting.CastingSpell  = item.Item.BackingAction;
Casting.SpellTarget   = target;
Casting.CastingHeal   = true;
Casting.SpellCastTime = TimeSpan.FromMilliseconds(8000);  // NOT AdjustedCastTime - see below
Casting.CastingTime.Restart();
return true;
```

Two traps in that:

- **Don't use `BackingAction.AdjustedCastTime`.** It reports `0` for Phoenix Down despite the visible 8s cast bar. A literal `8000` is uglier but honest — a `0` makes `CheckForSuccessfulCast`'s `buffer` maths meaningless, and since `UseAdvancedSpellHistory2` defaults true, the `AdjustedCastTime == 0 && Cooldown == 0` branch returns early and leaves `CastingTime` running. (`Range` and `AdjustedCooldown` off `BackingAction` are both correct — 15y and 360s — it's only the cast time that lies.)
- **`item.UseItem(target)` returns `bool` and it's currently discarded**, so `Execute()` reports success even when the use failed. The snippet above acts on it.

cheesegoldfish (commented):
This is a well-built PR and I want it in — Phoenix Down as a last-resort revive is very much on-topic for Magitek. Settings, XAML, and all three i18n files including the hand-maintained Designer are present and correct, which is rarer than it should be. Builds clean with no new warnings.

A few things I checked against the running client so they don't need re-litigating: item 4570 is Phoenix Down, `PhoenixDownRange = 15f` matches the item's own `BackingAction.Range`, and the alliance-scope worry I had (could this revive strangers?) is a non-issue — `SwitchCastableToAlliance` restores party scope before this runs.

Also, you were right and Codex was wrong on the reserve default: `Unique = false, StackSize = 999`. Settled, no need to revisit.

Three changes I'd like, all in `PhoenixDown.cs` and all inline: the 8-second blocking hold should use the existing `Casting` machinery, the two inventory scans should sit behind a cheap check, and the corpse range filter should use raw `Distance` per the AGENTS.md exception.

One other thing, on your retraction above — I think you may have withdrawn good evidence. `DutyManager.InInstance` was in your first commit, so it gated those casts, which means you *were* in an instance when they fired. Field operations (Eureka/Bozja/Occult Crescent) are instanced content that's also full of unrelated players running Sprint and Peloton, which matches what you saw in the combat log. Worth re-checking which zone that actually was before concluding the revive path has never worked — it may well have.

Magitek/Logic/Roles/PhoenixDown.cs
HomoYouDidnt: Done as specified: `UsePhoenixDown` → `DeadAllies.Count` → `InInstance`, then one bag walk that answers both the existence and the reserve questions. The comment carries your reasoning about why only `.Count` is safe to front-load, so the ordering doesn't get "simplified" back later.

Magitek/Logic/Roles/PhoenixDown.cs
HomoYouDidnt: Switched to raw `Distance` with the Healer.cs reasoning quoted at the site, and added `InLineOfSight()` to match. You were right to press on the CanUse/LoS claim — I couldn't verify it, so the comment no longer asserts it and the explicit filter carries it instead.

Magitek/Logic/Roles/PhoenixDown.cs
HomoYouDidnt: Adopted, and wiring it up surfaced three things worth your eyes since two touch `Casting.cs`:

1. **The per-job `NeedToInterruptCast` checks would have cancelled every tracked revive.** They cancel any cast whose target is dead unless it's that job's own raise spell — and a revive's target is dead by definition. BLM and BLU don't even carry a raise exemption, so on those jobs the registered cast died on its first tracked pulse and `Execute()` re-fired next pulse: a start/cancel loop that could never finish, on exactly the raise-less jobs this item is for. Rather than carving an exception into all seven job checks, there's one exemption in `TrackSpellCast` before the job dispatch, behind a new `Casting.CastingRevive` flag (set at registration, cleared everywhere `CastingTime` stops). The target-validity cancels above it still run — a despawned corpse still cancels.

2. **The literal 8000 alone didn't clear the advanced-history early return** — that check reads `CastingSpell.AdjustedCastTime`, which is still 0 for the BackingAction, not the `SpellCastTime` I set. Without an exemption it returns with the timer still running and the flag latched. `!CastingRevive` is now part of that condition so the revive falls through to the buffer maths the literal was set up for.

3. **Kept the `IsCasting` bridge from `DoAction` before registering.** Registering into the gap between `UseItem` and the cast bar lets one pulse see `!IsCasting` with the timer running, which `CheckForSuccessfulCast` reads as a failed cast — tracking dropped, and the walk-away bug comes straight back. It's the same wait your cited pattern does, not the 8-second hold.

`UseItem`'s bool is acted on, and the method stayed honestly async since a real await remains.

HomoYouDidnt:
You were right about the retraction, and checking it properly turned up a story none of the three versions had.

My error was a timezone slip: I checked the combat log at the RB timestamp raw, but RB logs are UTC and the combat log is local — four hours off. At the correctly converted times the character was in **the Clyteum**, across two separate runs of it, and Alphinaud (SGE) and G'raha Tia (BLM) were added as combatants at each zone-in — Duty Support, jobs matching the two log lines exactly. The "public zone full of strangers" was whatever was happening four hours later.

The selection evidence is actually better than my original comment claimed: Alphinaud died at 00:49:18 and was selected at 00:49:28 — exactly the configured 10-second delay — and G'raha at 01:18:17/01:18:27. Real, fresh corpses in a real instance. The stale-list mechanism in my retraction was wrong too.

But the original "demonstrably works" doesn't come back, for a new reason: both NPCs were up and casting again 3–5 seconds after the use began — inside the 8-second cast, which can't land before it finishes. The duty's own script revived them before the item could. So the selection-and-use path is field-proven; an end-to-end item revive still isn't, in any log I hold. The honest test plan remains a player party with a raise-less comp.

Magitek/Utilities/Casting.cs
```diff
@@ -102,6 +107,15 @@ public static async Task<bool> TrackSpellCast()
...
             }
 
+            // A revive (Phoenix Down) targets a body that is dead by definition, and the job checks
+            // below cancel any cast on a dead target unless it is that job's own raise spell — so every
+            // one of them would kill the revive on its first tracked pulse. None of their other cancels
+            // can apply either (each is keyed to the job's own spells), so exempt the revive here once
+            // rather than carving an exception into all seven jobs. The target-validity checks above
+            // still run: a despawned corpse still cancels.
+            if (CastingRevive)
+                return true;
```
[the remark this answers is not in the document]
HomoYouDidnt: Right, and this one's on me — the round that added the exemption explicitly considered this case and called it "the same trade as before," which missed that cancelling is strictly better: the item is saved and the player is free ~5 seconds sooner. The exemption now keeps the one cancel that is meaningful for a revive, in the same shape every healer applies to its own raise: target has the Raise aura or is back above 0 HP → cancel. Fixed in 28b171b3.

Magitek/Utilities/Casting.cs
```diff
@@ -22,6 +22,11 @@ internal static class Casting
     {
         #region Variables
         public static bool CastingHeal;
+        // True while the tracked cast is a revive (Phoenix Down). The healer NeedToInterruptCast
+        // checks cancel any cast whose target is dead unless it is that job's own raise spell —
+        // and a revive's target is dead by definition, so they need this to tell the two apart.
+        // Cleared wherever CastingTime stops, so it is true exactly while the revive is tracked.
+        public static bool CastingRevive;
```
cheesegoldfish, line 32: The lifecycle comment says "Cleared wherever `CastingTime` stops", and that's true of the three places it stops — but `CastingTime` also gets **restarted** in four places, and three of those don't clear the flag:

```
set:          PhoenixDown.cs:122
cleared:      Casting.cs:219 (CancelCast), 240 (CheckForSuccessfulCast early), 265 (main)
CastingTime.Restart():
              PhoenixDown.cs:124                       <- sets the flag first, fine
              SpellDataExtensions.cs:73, 320, 370      <- normal casts, flag not cleared
```

Those three are exactly where `Casting.CastingHeal = false;` already sits — the sibling flag is cleared there and this one isn't.

The window is narrow but reachable. While a revive is tracked, `TrackSpellCast()` returns `true`, which short-circuits `CheckForSuccessfulCast()` — so cleanup never runs for the whole 8-second cast. If the routine is stopped, the zone changes, or the player dies in that window, `CastingTime` is left running with `CastingRevive` still set. The next ordinary cast calls `CastingTime.Restart()` without clearing it, and `TrackSpellCast` then takes the revive branch for that cast — bypassing that job's entire `NeedToInterruptCast`.

One line in each of the three registration paths, next to the existing `CastingHeal` clear:

```csharp
Casting.CastingHeal = false;
Casting.CastingRevive = false;
```

Low probability, but `Casting` is shared by all 22 jobs and a latched exemption is invisible when it does happen — worth closing rather than reasoning about.

cheesegoldfish (commented):
All three of my previous comments are addressed, and the two things you found while wiring up the `Casting` registration were both real — I'd missed them.

You're right that `SpellCastTime = 8000` alone doesn't clear the advanced-history early return; that check reads `CastingSpell.AdjustedCastTime`, which stays 0 for the BackingAction regardless of what we set. My comment was incomplete and following it literally would have returned with the timer still running.

The `NeedToInterruptCast` problem is the bigger one and I didn't see it at all: a revive's target is dead by definition, so every one of those checks would have cancelled the cast on its first tracked pulse — and on BLM/BLU, which carry no raise exemption and are exactly the raise-less jobs this feature exists for, it would have start/cancel looped indefinitely. My suggested fix would have made things worse than the blocking wait it replaced. Solving it with one exemption ahead of the job dispatch rather than seven per-job carve-outs is the right shape, and keeping the someone-else-raised-them cancel is right too.

Build is clean — 0 errors, 60 warnings, identical to master, nothing naming the new files.

One thing left, inline, and it's a one-liner. After that I'm happy to merge this.

Magitek/Utilities/Casting.cs
HomoYouDidnt: Closed in b47b58e5 exactly as you laid it out — one clear beside each of the three existing `CastingHeal` writes, and the lifecycle comment now names the registration sites instead of overclaiming about stops. You're right that it's the invisible kind when it fires; belt-and-braces is the correct trade here.

PR #222, opened 2026-08-14
Dawntrail updates

lokibunny:
SGE - Fix Eukrasian Dyskrasia failing to target self. 
SGE - Optimize Psyche to cast on bosses regardless of AoE settings. 
SGE - Fix Eukrasian Prognosis II causing Pepsis and fight logic to stall.
SGE - Fix Philosophia aura tracking.

SCH - Implement dynamic spell handling for Concitation upgrade. 
SCH - Fix Seraphism stance causing Recitation and Emergency Tactics to hang.
SCH - Add safety checks to Baneful Impaction targeting.

AST - Overhaul Play logic to hold defensive cards appropriately. 
AST - Fix Helios Conjunction failing to cast at level 96.

WHM - Fix Presence of Mind and Glare IV deadlock. 
WHM - Optimize Divine Caress to not overwrite Temperance. 
WHM - Add safety checks to Glare IV targeting.

BLM - Update rotation logic for Dawntrail mechanics and fix phase stalling


Magitek/Logic/WhiteMage/Buff.cs
```diff
@@ -43,9 +43,12 @@ public static async Task<bool> PresenceOfMind()
             if (!Core.Me.InCombat)
                 return false;
 
+            // Dawntrail Fix: Do not use IsKnownAndReady() on Glare IV here. 
+            // Glare IV requires the Sacred Sight aura to be "Ready", which PoM grants. 
+            // Using IsKnownAndReady() creates a deadlock where PoM never fires.
             if (WhiteMageSettings.Instance.PresenceOfMindForGlareIV
                 && Spells.PresenceofMind.IsKnownAndReady()
-                && Spells.GlareIV.IsKnownAndReady())
+                && Spells.GlareIV.IsKnown())
```
cheesegoldfish, line 51: The fix is right but the stated reason isn't, and I want the reason recorded correctly so it doesn't get "corrected" back later.

`IsKnownAndReady()` is `IsKnown() && IsReady()` (`Extensions/SpellDataExtensions.cs:446`) — it reads `ActionManager.HasSpell` and the cooldown. Sacred Sight never enters into it. Verified against the running client on a level 96 WHM with no Sacred Sight up:

```
HasSacredSight      = false
GlareIV_HasSpell    = true
GlareIV_Cooldown_ms = 0     -> IsKnownAndReady() == true
```

So there is no Sacred Sight deadlock.

The real cause is that Glare IV is a GCD, so its `Cooldown` is the GCD timer. Presence of Mind is an oGCD we want to weave, and weaving happens *while the GCD is rolling* — exactly when `GlareIV.IsKnownAndReady()` is false. PoM only had a window in the instant the GCD was already up, which is when we'd be casting a GCD instead. That's why it gets skipped.

Dropping to `IsKnown()` does resolve it, so please keep the change — but replace the comment with the GCD explanation.

Magitek/Logic/WhiteMage/Heal.cs
```diff
@@ -687,6 +687,18 @@ public static async Task<bool> DivineCaress()
...
+            // Dawntrail Optimization: Prevent overwriting Temperance's mitigation window with the shield.
+            // Hold Divine Caress unless Temperance has fallen off, someone is critical, or the Grace buff is expiring.
+            bool temperanceActive = Core.Me.HasAura(Auras.Temperance);
+
+            // If we DO NOT have the aura with at least 5 seconds left, it is expiring.
+            bool graceExpiring = !Core.Me.HasAura(Auras.DivineGrace, true, 5000);
+
+            bool emergencyHealNeeded = Group.CastableAlliesWithin30.Any(r => r.CurrentHealthPercent <= WhiteMageSettings.Instance.AoEHealHealthPercent);
+
+            if (temperanceActive && !graceExpiring && !emergencyHealNeeded)
```
cheesegoldfish: Can you back this one up? Divine Caress does not overwrite Temperance — they're separate statuses and they stack. I can't reproduce the behaviour this is guarding against.

As written this holds Divine Caress whenever Temperance is up, which is most of the window we'd actually want to use it in, and it reuses `AoEHealHealthPercent` as an "emergency" threshold for an unrelated purpose.

What's the observed problem this fixes? If it's a sequencing preference rather than a bug, it needs to be a setting rather than hardcoded behaviour.

Magitek/Logic/Sage/AoE.cs
```diff
@@ -217,10 +218,14 @@ public static async Task<bool> Psyche()
             if (target == null)
                 return false;
 
-            if (Combat.Enemies.Count(r => r.Distance(target) <= Spells.Psyche.Radius) < SageSettings.Instance.PsycheAoEEnemies)
+            if (!Spells.Psyche.IsKnownAndReady())
                 return false;
 
-            if (!Spells.Psyche.IsKnown())
+            // Dawntrail Fix: Psyche is a pure DPS gain. Use it on cooldown against bosses, or respect AoE thresholds for trash.
+            bool isBoss = target.IsBoss();
```
cheesegoldfish: `IsBoss()` isn't reliable enough to gate a damage cooldown on — it isn't always up to date — and this now silently ignores `PsycheAoEEnemies` for anything it flags as a boss.

Psyche is always a DPS gain, so if the intent is "use it on cooldown", the honest version is to let people set `PsycheAoEEnemies` to 1 (which is what I run) or add an explicit "always use Psyche" toggle. Please drop the `IsBoss` branch and do it through the setting.

Magitek/Logic/Sage/Buff.cs
```diff
@@ -235,8 +235,12 @@ public static async Task<bool> Philosophia()
             if (Spells.Philosophia.Cooldown != TimeSpan.Zero)
                 return false;
 
-            var targets = Group.CastableAlliesWithin30.Where(r => r.CurrentHealthPercent < SageSettings.Instance.PhilosophiaHealthPercent
-                                                                  && !r.HasAura(Auras.Eudaimonia));
+            // Dawntrail Fix: Philosophia (Eudaimonia) applies to the Sage, not the party members.
```
cheesegoldfish: This comment is incorrect. Philosophia's tooltip reads:

> Increases healing magic potency by 20%. Duration: 20s. Additional Effect: Grants self and nearby party members the effect of Eudaimonia, restoring HP after landing spells.

Eudaimonia goes to self *and* nearby party members, so removing `!r.HasAura(Auras.Eudaimonia)` from the party filter drops a check that was doing real work.

Separately, the function already returns early on `Spells.Philosophia.Cooldown != TimeSpan.Zero`, so it can't re-cast while the buff is up anyway. What was the observed failure here?

Magitek/Logic/Sage/Heal.cs
```diff
@@ -448,10 +448,13 @@ private static async Task<bool> UseEukrasianPrognosisIfNeeded(int NeedShields, S
 
             if (needPrognosis)
             {
-                if (!await UseEukrasia(Spells.EukrasianPrognosis.Id))
+                // Dawntrail Fix: Dynamically upgrade spell ID to prevent Pepsis hanging at Lv. 96+
+                var spell = Spells.EukrasianPrognosisII.IsKnown() ? Spells.EukrasianPrognosisII : Spells.EukrasianPrognosis;
```
cheesegoldfish: This isn't necessary — RebornBuddy already proxies straight-upgrade actions. Verified against the running client on a level 100 Scholar:

```
Succor (186) -> HasSpell = true, GetMaskedAction -> Concitation (37013), CanCast(self) = true
```

Confirmed the same mechanic covers conditional replacements too — see my comment on `Scholar/Heal.cs`. The same applies to Eukrasian Prognosis; the `ForcePrognosis` toggle already casts Prognosis II correctly on master with no dynamic ID selection at all.

This file also already had an `EukrasianPrognosisII.IsKnown()` branch at line 265, so this adds a second, differently-styled copy of a pattern the file already contained.

Please drop this and the matching change in `HealFightLogic.cs:106`. If a spell genuinely needs resolving at runtime, use `.Masked()` — that's the established pattern here.

Magitek/Logic/Scholar/Heal.cs
```diff
@@ -111,12 +111,14 @@ public static async Task<bool> ForceSuccor()
             if (!ScholarSettings.Instance.ForceSuccor)
                 return false;
 
-            if (!await Spells.Succor.Cast(Core.Me)) return false;
+            var succorSpell = Spells.Concitation.IsKnown() ? Spells.Concitation : Spells.Succor;
```
cheesegoldfish, line 114: Tested this against the running client rather than reasoning about it, and the premise doesn't hold — RebornBuddy already proxies both upgrades.

Level 100 Scholar, Seraphism **down**:

```
GetMaskedAction(186 Succor)    -> Concitation (37013)
GetMaskedAction(185 Adloquium) -> Adloquium   (185)
CanCast(185/186, self)         -> true / true
```

Same character, Seraphism **up**:

```
GetMaskedAction(185 Adloquium) -> Manifestation (37015)
GetMaskedAction(186 Succor)    -> Accession     (37016)
CanCast(185, self)             -> true
CanCast(186, self)             -> true
```

`ActionManager.CanCast(Spells.Adloquium.Id, Core.Me)` returns **true** under Seraphism — that is the exact call this PR says hangs. `Spells.Succor.Cast()` already casts Concitation, and Accession under Seraphism. Every ternary added here and in the other three Scholar files is dead code.

It also introduces a regression. `Casting.CastingSpell` is set to *the SpellData passed to `.Cast()`* (`Extensions/SpellDataExtensions.cs:69`), not the masked result. `Utilities/Routines/Scholar.cs:47` depends on that:

```csharp
if (Casting.CastingSpell == Spells.Succor || Casting.CastingSpell == Spells.Adloquium)
    return false;   // exempt our shields from the interrupt-healing check below
```

Once these call sites start passing `Concitation` / `Accession` / `Manifestation`, that exemption stops matching and those casts fall through to `InterruptHealing` — so at level 96+ the bot can cancel its own AoE shields mid-cast whenever party health recovers.

Please revert the Scholar spell-swapping entirely. If something ever does need resolving at runtime, `.Masked()` is the pattern we use here, and it should be one resolved reference rather than a ternary at each call site — but nothing in this file needs it.

Magitek/Logic/BlackMage/SingleTarget.cs
```diff
@@ -333,7 +295,8 @@ public static async Task<bool> Paradox()
             if (Casting.LastSpellWas(Spells.Blizzard3))
                 return false;
 
-            if (AstralStacks != 3 && UmbralStacks != 3)
+            // Dawntrail check: Paradox is strictly available in Astral Fire
```
cheesegoldfish: This is wrong, and it's a rotation regression. Paradox is used in both Astral Fire and Umbral Ice — it's the standard transition out of ice (Blizzard III -> Blizzard IV -> Paradox -> Fire III).

Icy Veins' Dawntrail rotation page: "Paradox replaces Fire and Blizzard on your hotbars when conditions are met. It should be used in both Astral Fire and Umbral Ice." I've also confirmed in game that Paradox comes up on every swap between fire and ice, not just in fire.

`AstralStacks < 3` removes it from the ice phase entirely. Please restore the original `AstralStacks != 3 && UmbralStacks != 3` gate.

Magitek/Logic/BlackMage/Aoe.cs
```diff
@@ -113,22 +88,19 @@ public static async Task<bool> Freeze()
...
-            //If flarestar is ready, cast it
             if (AstralSoulStacks == 6)
                 return false;
 
-            //Can only use in Umbral Ice
             if (UmbralStacks != 3)
                 return false;
 
-            if (Core.Me.CurrentMana == 10000)
+            if (UmbralHearts == 3)
```
cheesegoldfish, line 192: This changes what the check means. `CurrentMana == 10000` was "stay in ice until MP is actually restored", which is the entire point of the ice phase. `UmbralHearts == 3` is satisfied by a single Freeze, since Freeze grants all three hearts at once.

Concretely: enter Umbral Ice III with hearts already banked and low MP, and Freeze is now blocked immediately, so we never regenerate. Nothing else in the AoE list covers it — `Blizzard2` returns false on `UmbralStacks == 3`, and `Flare` returns false on `UmbralStacks > 0`.

Having 3 hearts does not mean we should leave the ice AoE rotation. What was this fixing?

Magitek/Logic/BlackMage/Buff.cs
```diff
@@ -298,15 +213,14 @@ public static async Task<bool> Amplifier()
             if (!BlackMageSettings.Instance.Amplifier)
                 return false;
 
-            if (ActionResourceManager.BlackMage.PolyglotCount > 2)
+            if (!Core.Me.InCombat)
                 return false;
 
-            if (!Core.Me.InCombat)
+            int maxPolyglot = Core.Me.ClassLevel >= 98 ? 3 : (Core.Me.ClassLevel >= 80 ? 2 : 1);
```
cheesegoldfish, line 296: This duplicates `Magitek.Utilities.Routines.BlackMage.MaxPolyglotCount`, which already exists and which this file already imports as `BlackMageRoutine`. Please use it instead:

```csharp
if (ActionResourceManager.BlackMage.PolyglotCount >= BlackMageRoutine.MaxPolyglotCount)
    return false;
```

The two have already drifted — the shared one returns 0 below level 70, this one returns 1. Per AGENTS.md, shared level/trait calculations belong in `Utilities/Routines/<Job>.cs` so every file gets the same answer.

Magitek/Logic/Astrologian/Cards.cs
```diff
@@ -12,97 +12,65 @@ namespace Magitek.Logic.Astrologian
...
+                return false;
 
-            //if (await RedrawOrDrawAgain(drawnCard))
-            //    return true;
+            // Dump cards if next draw is within 5 seconds so we don't overcap
+            // Astral and Umbral share a recast timer
+            bool forceDump = false;
+            if (Core.Me.InCombat)
+            {
+                if (Spells.AstralDraw.IsKnown() && Spells.AstralDraw.Cooldown.TotalSeconds < 5) forceDump = true;
```
cheesegoldfish: Two problems here.

The thresholds are hardcoded. The `5` seconds here and `CurrentHealthPercent <= 80` below both need to be real settings with `[Setting]` + `[DefaultValue]`, a control in the AST view, and strings in both `Resources.resx` and `Resources.zh-CN.resx`. See AGENTS.md, "Configuration and Hardcoded Values".

The hold logic also doesn't hold for long. `Draw()` only fires when every card is `None`, so while we're sitting on defensive cards the draw never goes out and its cooldown parks at 0. `Cooldown.TotalSeconds < 5` is then permanently true, `forceDump` is stuck on, and defensive cards get dumped on a full-health tank — which is the behaviour this PR says it's fixing. It only really holds for the first draw cycle of a fight.

Magitek/Logic/BlackMage/Aoe.cs
```diff
@@ -20,41 +19,27 @@ public static async Task<bool> Foul()
...
                 return false;
 
-            // If we're moving in combat
             if (MovementManager.IsMoving)
             {
-                // HARDCODED: Level 80 is when Foul becomes instant cast via trait
-                // This is a trait check, not a spell availability check
-                // If we have instant cast Foul (level 80+) and have Swiftcast/Triplecast,
-                // don't cast Foul - use procs on other spells instead
                 if (Core.Me.ClassLevel >= 80 && (Core.Me.HasAura(Auras.Swiftcast) || Core.Me.HasAura(Auras.Triplecast)))
```
cheesegoldfish, line 41: Please restore the comments removed here. The block above this line explained why level 80 is checked directly — it's the trait that makes Foul instant, not a spell availability check. That's exactly the kind of comment that has to stay, because without it the next person "fixes" it into an `IsKnown()` check.

Magitek/Logic/BlackMage/SingleTarget.cs
```diff
@@ -76,52 +73,44 @@ public static async Task<bool> Despair()
...
                 return false;
 
-            //only use in astral fire
             if (UmbralStacks > 0)
                 return false;
 
-            // FFXIV MP costs (patch 7.x): Fire base cost 800 MP, doubled to 1600 in Astral Fire.
-            // Spells.Fire.Cost is NOT reliably dynamic for AF stance; hardcode known game values.
-            // If SQEX changes Fire's MP cost in a future patch, update these constants.
             if (AstralStacks > 0 && Core.Me.CurrentMana < 1600)
```
cheesegoldfish, line 124: The comment removed from above this line documented why 1600 is hardcoded:

```csharp
// FFXIV MP costs (patch 7.x): Fire base cost 800 MP, doubled to 1600 in Astral Fire.
// Spells.Fire.Cost is NOT reliably dynamic for AF stance; hardcode known game values.
// If SQEX changes Fire's MP cost in a future patch, update these constants.
```

Please put it back. A bare `1600` with no explanation is exactly what that comment existed to prevent.

Magitek/Logic/Astrologian/Cards.cs
```diff
@@ -129,11 +97,9 @@ public static async Task<bool> Draw()
 
         private static GameObject Tank()
         {
-            //Get party size
             int partySize = Group.CastableAlliesWithin30.Count();
```
cheesegoldfish, line 84: `//If in light party, allow ally to have more than one card aura.` was removed from this function and the two below it. Please restore all three — they explain why the `partySize <= 4` branch exists, which isn't obvious from the code.

Magitek/Logic/Astrologian/Heals.cs
```diff
@@ -453,26 +453,26 @@ private static async Task<bool> AspectHeliosInsteadOfDiurnalBenefic()
             if (AstrologianSettings.Instance.DisableSingleHealWhenNeedAoeHealing && NeedAoEHealing())
                 return false;
 
-            if (AstrologianSettings.Instance.DisableSingleHealWhenNeedAoeHealing && NeedAoEHealing())
-                return false;
+            var spell = Spells.HeliosConjunction.IsKnown() ? Spells.HeliosConjunction : Spells.AspectedHelios;
```
cheesegoldfish, line 456: Same problem as the Scholar changes, and it lands somewhere that's actually read.

`Casting.CastingSpell` stores the SpellData passed to `.Cast()`, not the masked result. `Utilities/Routines/Astrologian.cs:57` depends on that:

```csharp
if (Casting.CastingSpell == Spells.AspectedHelios)   // inside the InterruptHealing block
```

Once this casts via `spell` (Helios Conjunction at 96+), that check stops matching and we lose the ability to cancel an Aspected Helios / Helios Conjunction that's no longer needed.

The aura fix in this function is right and worth keeping — at 96+ the cast applies `HeliosConjunction` (3894), not `AspectedHelios` (836), so adding `!r.HasMyAura(Auras.HeliosConjunction)` to the filter fixes a real re-cast loop. But the spell swapping isn't needed: RebornBuddy masks the action for you (verified live — details in my `Scholar/Heal.cs` comment). Please keep the aura change and drop the `var spell = ...` selection.

cheesegoldfish (changes requested):
Thanks for putting time into this. There's some good work in here, but I need a fair bit changed before it can go in.

**On scope — for next time rather than this one.** This is 19 files across 5 jobs plus two shared utility files. I'd rather not send you back to re-cut it now, so let's work this one as-is. But for future PRs, could you keep them to roughly one job each and split shared-file changes (`Utilities/Group.cs`, `Utilities/Tracking.cs` here) into their own? It's purely about what I can verify — a single-job PR I can test and merge in an evening, where one this size I end up sitting on. See CONTRIBUTING.md § PR Scope.

**Please stop removing comments.** A large share of the 458 deletions is existing comments being stripped, including ones documenting why specific magic numbers and level checks exist. AGENTS.md is explicit: don't change existing comments unless they're factually wrong, and don't remove commented-out code you didn't write. Several of the deleted ones are load-bearing — I've flagged a few inline, but please restore all of them, not just those. Same for the stripped BOMs and trailing newlines; that's diff noise for no benefit.

**Several of the claims don't hold up.** I checked these against the running client and against official sources rather than by eye; details are inline:

- Glare IV does not gate `IsKnownAndReady()` on Sacred Sight. Real bug, wrong cause — the GCD timer was what blocked the weave.
- Divine Caress does not overwrite Temperance; they stack.
- Philosophia grants Eudaimonia to self *and* nearby party members, per its tooltip.
- Paradox is used in Umbral Ice as well as Astral Fire, so that change is a rotation regression.
- **RebornBuddy already proxies upgraded actions, including Seraphism.** Verified live on a level 100 Scholar: with Seraphism up, `Adloquium` masks to `Manifestation`, `Succor` masks to `Accession`, and `CanCast` returns true for both. The dynamic spell-ID work across Scholar and Sage isn't fixing anything, and it introduces a regression in `Utilities/Routines/Scholar.cs:47` — detail in that comment.

That last point covers most of the Scholar and Sage changes here, so for those two jobs I'd rather see the spell-swapping reverted than revised.

A couple of things are genuinely good and I'd take right away: the Eukrasian Dyskrasia self-target fix is correct (Range 0, Radius 5 — verified), and the HashSet change in `Tracking.cs` is a real improvement.

I won't merge with unresolved comments, so please work through the inline ones as you go.

Magitek/Logic/BlackMage/Aoe.cs
```diff
@@ -20,76 +19,116 @@ public static async Task<bool> Foul()
...
+            // In Umbral Ice with 3 Umbral Hearts -> Transpose to Astral Fire
+            if (UmbralStacks > 0 && UmbralHearts == 3)
+            {
+                return await Spells.Transpose.Cast(Core.Me);
+            }
+
+            return false;
+        }
+
+        public static async Task<bool> UseAoeEther()
```
cheesegoldfish, line 91: This method can never do anything. The `if` that guarded this `return false;` was commented out along with the diagnostic logs above it, so the bare `return false;` is now unconditional and everything below is dead:

```csharp
// Diagnostic log
// Logger.WriteInfo($"[BLM-AoE-Ether] Checking: ...");

    return false;                     // <- no longer attached to any condition

if (Core.Me.CurrentMana >= 800)       // <- unreachable
    return false;
```

The compiler agrees — `Magitek/Logic/BlackMage/Aoe.cs(93,13): warning CS0162: Unreachable code detected`. Build warnings go 60 -> 64 with this commit.

It's wired in at `Rotations/BlackMage.cs:80` and ships with the new `UseEtherInAoe` setting, a checkbox in `Views/UserControls/BlackMage/Aoe.xaml`, and strings in both `Resources.resx` and `Resources.zh-CN.resx`. So a user can enable "Use Ethers in AoE" and nothing will ever happen.

The intent looks fine — ether for an extra Flare when Manafont is down. It just needs the guard restored.

Magitek/Logic/BlackMage/Buff.cs
```diff
@@ -150,21 +123,16 @@ public static async Task<bool> UmbralSoul()
...
-
-            // Do not Umbral Soul unless we have 1 umbral stack
-            if (UmbralStacks == 0)
+            // Do not use in combat if we have a target (use Aoe.Freeze instead)
+            if (Core.Me.InCombat && Core.Me.HasTarget)
                 return false;
 
-            if (Core.Me.CurrentTarget != null)
-                return false;
+            if (UmbralStacks == 0)
```
cheesegoldfish, line 188: The `return false;` under this `if` was removed, which leaves it bodiless — so it binds to the *next* statement instead:

```csharp
if (UmbralStacks == 0)
    if (UmbralStacks < 3 || UmbralHearts < 3)
        return await Spells.UmbralSoul.Cast(Core.Me);
```

That inverts the guard. Umbral Soul now only fires when `UmbralStacks == 0`, but the ability can only be executed under Umbral Ice, so it can never fire at all.

This one compiles clean and produces no warning, which is what makes it nasty — it will just look like Umbral Soul quietly stopped working.

`PreCombatUmbralSoul()` just below was restructured the same way but kept its `return false;`, so that one is fine.

Magitek/Utilities/Ether.cs
```diff
@@ -16,10 +16,16 @@ public static async Task<bool> UseEther(int etherItemId)
             var etherItem = InventoryManager.FilledSlots.FirstOrDefault(s => s.RawItemId == etherItemId);
 
             if (etherItem == null)
+            {
+                Logger.WriteInfo($"[BLM-AoE-Ether] Ether item not found: {etherItemId}");
```
cheesegoldfish, line 23: These two `Logger.WriteInfo` calls aren't gated behind a debug setting, and `UseEther` is called from a per-pulse rotation path. Right now they're masked by the dead code in `UseAoeEther()`, but as soon as that's fixed this will write two lines per pulse for anyone who doesn't happen to be carrying ethers.

Either gate them on `BaseSettings.Instance.DebugFightLogic` (or a similar debug flag) the way the rest of the codebase does, or drop them once the feature is working.

Magitek/Rotations/BlackMage.cs
```diff
@@ -62,8 +65,9 @@ public static async Task<bool> Combat()
             if (await Buff.Triplecast()) return true;
             if (await Buff.LeyLines()) return true;
             if (await Buff.Retrace()) return true;
+            if (await CombatBuff()) return true;
```
cheesegoldfish: `CombatBuff` is one of the `IRotation` methods that `RotationManager` already dispatches on its own via reflection (`Utilities/Managers/RotationManager.cs`), so calling it from inside `Combat()` runs the potion logic twice per pulse rather than once.

If the intent is "use a potion during combat on bosses", that belongs in the `CombatBuff()` body alone and the manager will call it — no call needed from `Combat()`.

Minor, in the same line: `global::Magitek.Utilities.Combat.IsBoss()` is working around a name collision with `ff14bot`'s `Combat`. A `using` alias at the top of the file (the codebase already does this, e.g. `using BlackMageRoutine = Magitek.Utilities.Routines.BlackMage;`) reads better than `global::` at the call site.

Magitek/Logic/BlackMage/SingleTarget.cs
```diff
@@ -327,23 +296,33 @@ public static async Task<bool> Paradox()
...
 
             if (Spells.ManaFont.IsKnownAndReady())
                 return false;
 
-            if (Spells.Fire4.IsKnownAndReadyAndCastableAtTarget() && Spells.ManaFont.Cooldown.TotalMilliseconds >= 70000)
+            // In Astral Fire, cast Paradox after 3x Fire 4 (AstralSoulStacks >= 3) or when moving
+            if (MovementManager.IsMoving)
+                return await Spells.Paradox.Cast(Core.Me.CurrentTarget);
+
+            if (AstralSoulStacks < 3 && Spells.Fire4.IsKnown())
```
cheesegoldfish, line 383: This part is correct and worth keeping — Icy Veins' Astral Fire phase is `Fire IV x3 -> Paradox -> Fire IV x3 -> Flare Star -> Despair`, and since Fire IV grants Astral Soul, `AstralSoulStacks >= 3` puts Paradox in exactly the right slot. Better than the `ManaFont.Cooldown >= 70000` check it replaces.

The problem is the gate ~10 lines above it, which is unchanged from my earlier comment:

```csharp
// Dawntrail check: Paradox is strictly available in Astral Fire
if (AstralStacks < 3)
    return false;
```

Paradox is used in Umbral Ice too — it's the standard transition out of ice (`Blizzard III -> Blizzard IV -> Paradox -> Fire III`), and Icy Veins states it directly: "Paradox replaces Fire and Blizzard on your hotbars when conditions are met. It should be used in both Astral Fire and Umbral Ice." I've also confirmed in game that it comes up on every swap between fire and ice.

So right now the Astral Fire half is correct and the Umbral Ice half is switched off. Restoring `AstralStacks != 3 && UmbralStacks != 3` should give you both.

Magitek/Logic/BlackMage/Aoe.cs
```diff
@@ -113,22 +152,19 @@ public static async Task<bool> Freeze()
...
             if (AstralSoulStacks == 6)
                 return false;
 
-            //Can only use in Umbral Ice
-            if (UmbralStacks != 3)
+            if (UmbralStacks == 0)
                 return false;
 
-            if (Core.Me.CurrentMana == 10000)
+            if (UmbralHearts == 3)
```
cheesegoldfish, line 192: Retracting my earlier comment on this method — **the change is correct and my objection was wrong.**

I said `UmbralHearts == 3` was the wrong stop condition and that `CurrentMana == 10000` should stay. That was based on reading Freeze in isolation, before `AoeTranspose()` existed. With Transpose in the loop:

- The AoE rotation now enters Umbral Ice via Transpose, which gives UI**1**, not UI3. The old `if (UmbralStacks != 3) return false;` would have blocked Freeze entirely in that loop, so `UmbralStacks == 0` is required here.
- What gates the two Flares is **Umbral Hearts**, not full MP — hearts negate the Astral Fire cost increase. So `UmbralHearts == 3` is the right stop, and there's no MP deadlock to worry about.

The two changes are a package and I reviewed half of it. Apologies for the noise on that one.

cheesegoldfish (commented):
Reviewed the new commit on its own (diffed against `4e19639f`, so this only covers the ~245 new lines, not a re-read of the whole PR).

Good news first, because it's most of the rotation work: I checked the BLM and DRG changes against the published rotations rather than just reading them, and the rotation *logic* largely holds up.

- **Paradox after 3x Fire IV is correct.** Icy Veins' Astral Fire phase is `Fire IV x3 -> Paradox -> Fire IV x3 -> Flare Star -> Despair`, so gating on `AstralSoulStacks >= 3` matches it. That's a real improvement on the old `ManaFont.Cooldown >= 70000` heuristic.
- **`AoeTranspose()` matches the guide's AoE loop** (`Freeze -> Foul/HT2 -> Transpose -> Flare x2 -> Flare Star -> ...`). The AoE rotation previously had no Transpose at all, so this is a genuine fix.
- **The Freeze change is right and I was wrong about it** — retraction inline on that method.
- **DRG Life Surge now targets Heavens' Thrust and Drakesbane**, which is what the guide calls for, and it drops Chaotic Spring, which the guide explicitly says to avoid. That fixes a pre-existing bug.
- **The ether IDs are a real fix.** `MegaEther` was 13638, which is actually Max-Ether. Verified against the client: 4557 = Mega-Ether, 13638 = Max-Ether, and nothing else referenced those constants.

The problem is the implementation rather than the rotation knowledge. There are three bugs inline below that mean parts of this cannot have been executed even once — `UseAoeEther()` returns unconditionally before doing anything (the compiler flags it, `CS0162`), and `UmbralSoul()` has a bodiless `if` that inverts its guard so it can never fire. Both ship as user-facing behaviour, and one of them has a new setting, a checkbox and two translations attached to it.

Worth building and loading the DLL once before the next push — that would have caught all three in a single pull.

Separately: the 14 comments from the previous review are still open, and this push adds Dragoon and Red Mage on top (19 files -> 33). The Paradox gate below is the clearest example of why that ordering hurts — the new Astral Fire logic is correct, but it's sitting on top of an unfixed regression, so the good change can't land.

Magitek/Logic/Dragoon/Buff.cs
```diff
@@ -56,43 +57,56 @@ public static async Task<bool> LifeSurge() // Crit +10%
...
+                    }
+                    else
+                    {
+                        // With 1 charge, save it for LanceCharge
+                        return false;
+                    }
+                }
+            }
+
+            // Prioritize highest-potency weaponskills: Heavens' Thrust / Full Thrust > Drakesbane > Fang & Claw / AoE
```
cheesegoldfish, line 83: This is a good change and I want to flag it as such, since most of my comments are negative.

Icy Veins: "you will almost always want to use it on either Heavens' Thrust or Drakesbane, as they have our highest potency among our Weaponskills with 460 base", and separately that avoiding Chaotic Spring makes sense because Life Surge doesn't buff DoTs.

Firing after `VorpalThrust`/`LanceBarrage` (into Heavens' Thrust) and after `WheelingThrust`/`FangAndClaw` (into Drakesbane) hits exactly those two, and it matches the guide's combo order. The code this replaces included `ChaoticSpring` in the condition, so this fixes a real pre-existing bug rather than just reshuffling it.

One small thing: `Spells.LifeSurge.Charges < 1.5` is a hardcoded threshold that changes behaviour, so per AGENTS.md it should be a `[Setting]` with a control and both resx entries — or, if you'd rather not add another knob, a comment explaining what 1.5 represents would do.

Magitek/Logic/WhiteMage/Buff.cs
lokibunny: Fixed. Kept the IsKnown() change but replaced the Sacred Sight comment with an accurate explanation of the GCD timing block as requested.

Magitek/Logic/WhiteMage/Heal.cs
lokibunny: Fixed. Removed the factual error about overwriting Temperance and removed the misuse of AoEHealHealthPercent as an emergency threshold. The logic now safely chains them without holding it hostage.

Magitek/Logic/Sage/AoE.cs
lokibunny: Dropped the IsBoss() branch entirely. Psyche now strictly respects the PsycheAoEEnemies setting as requested.

Magitek/Logic/Sage/Buff.cs
lokibunny: Good catch, tooltip verified. I've restored the !r.HasAura(Auras.Eudaimonia) filter and removed the incorrect comment.

Magitek/Logic/Sage/Heal.cs
lokibunny: Completely reverted across both Sage and Scholar. Dropped the var spell = ... selections and went back to relying on RebornBuddy's .Masked() native handling.

Magitek/Logic/Astrologian/Cards.cs
lokibunny: Reverted this entire block to keep the PR scope clean and avoid the hardcoded thresholds / soft-locks.

Magitek/Logic/BlackMage/Aoe.cs
lokibunny: Restored.

Magitek/Logic/BlackMage/SingleTarget.cs
lokibunny: Restored the explanation for the 1600 MP hardcode.

Magitek/Logic/Astrologian/Cards.cs
lokibunny: Restored the Light Party explanatory comments across all three targeting methods.

Magitek/Logic/Astrologian/Heals.cs
lokibunny: Reverted the dynamic var spell selection here too so we don't break the interrupt checks, but kept the !r.HasMyAura(Auras.HeliosConjunction) fix to prevent the recast loop.

Magitek/Logic/BlackMage/Aoe.cs
lokibunny: Fixed. Removed the orphaned return false; and properly wired UseAoeEther() into the AoE Finisher phase right after Flare Star.

Magitek/Logic/BlackMage/Buff.cs
lokibunny: Fixed. Restored the missing return false; so the statement binds correctly and Umbral Soul can actually fire again.

Magitek/Utilities/Ether.cs
lokibunny: Wrapped the ether not found warnings in a BaseSettings.Instance.DebugFightLogic check.

Magitek/Rotations/BlackMage.cs
lokibunny: Removed the duplicate CombatBuff() call from inside Combat().

Magitek/Logic/BlackMage/SingleTarget.cs
lokibunny: Fixed. Restored the gate to if (AstralStacks != 3 && UmbralStacks != 3) so Paradox properly fires as the transition piece in Umbral Ice as well.

Magitek/Logic/BlackMage/Aoe.cs
lokibunny: Fixed. Restored the gate to if (AstralStacks != 3 && UmbralStacks != 3) so Paradox properly fires as the transition piece in Umbral Ice as well.

Magitek/Logic/Dragoon/Buff.cs
lokibunny: Added a comment explaining that the 1.5 threshold is specifically to prevent overcapping charges before the next Lance Charge window, avoiding the need for a new UI toggle setting.

lokibunny:
Breaking up into smaller PRs

PR #229, opened 2026-08-17
Core - Fix crashes when a spell target dies during a cast

HomoYouDidnt:
## What this changes

When a spell's target dies and despawns mid-cast, the routine now cancels the cast cleanly instead of crashing the whole combat pulse. Single-file change to the shared cast-tracking method that runs every job's mid-cast interrupt checks.

## Why

Live crash: `ReadWriteMemoryException ... at addr: 1AC` thrown from the Black Mage interrupt check when its cast target died and despawned mid-cast (Occult Crescent critical encounter). The exception killed the combat pulse. Every caster/healer job's interrupt check reads the same `Casting.SpellTarget`, so any of them could hit it.

Root cause, visible in the log: in `TrackSpellCast`, the "Target is no Longer Targetable" branch cancels the cast but did not return, and `CancelCast` awaits across frames — so the per-job interrupt checks could run after the target had despawned and its memory was freed. A freed target keeps a non-null `SpellTarget` reference, so the existing null-conditional checks pass, and the first field read through the stale pointer throws. The log shows "Target is no Longer Targetable" followed 59 ms later by the crash, in the same pulse.

## Fix

- The untargetable branch (and the catch beside it) now `return true;` after cancelling — the cancel-then-fall-through was the observed crash path, and it also spent gambit bookkeeping on a dead cast.
- `CancelCast` stops the cast timer and clears the revive flag in a `finally`, so the "already cancelled this pulse" guard holds even when StopCasting or its wait throws mid-transition.
- Immediately before the revive check and the per-job interrupt switch: stop if the cast is no longer being tracked, then re-check target validity **inside** the try — `IsValid` reads a liveness stamp through the object's pointer and can itself throw once the target's memory is unmapped, so the re-check sits under the same catch as the reads it guards.
- The catch captures the exception and logs it (`[Casting] Interrupt checks failed, cancelling cast: ...`), and each cancel site emits a distinct message, so log ordering stays diagnosable and a job-check bug is not silently reported as an invalid target.
- The revive block and the job switch are unchanged — only re-indented by the try block. Easiest to review with whitespace changes hidden.

None remaining. One claim in the original version of this body was wrong and is corrected above: `IsValid` is not a purely managed liveness check — it short-circuits safely on a zeroed pointer, but otherwise reads a stamp through the pointer and can throw on unmapped memory. That is why the validity re-check lives inside the try.

## In-game test plan

1. Setup: any caster (BLM is easiest), fast-dying trash — open world or Occult Crescent groups.
2. Trigger: keep casting at targets that die mid-cast, repeatedly.
3. Expect: casts stop cleanly with the existing messages ("because HE'S DEAD, JIM!" / "Stopped Cast: Unit Died"), no `ReadWriteMemoryException` in the log, rotation continues normally.
4. Raise path: healer raise where the corpse is claimed by another raiser or released mid-cast — expect the existing "Stopped Resurrection" messages, unchanged.
5. Zone or duty transition mid-cast: cast should cancel without an "Interrupt checks failed" line; if that line ever appears, it now carries the underlying exception message.

Deployed locally during live Occult Crescent play after the change: mid-cast target deaths produce clean interrupts and no exceptions.

### DIFF
--- a/Magitek/Utilities/Casting.cs
+++ b/Magitek/Utilities/Casting.cs
@@ -96,12 +96,14 @@ namespace Magitek.Utilities
                 if (!SpellTarget.IsTargetable)
                 {
                     await CancelCast("Target is no Longer Targetable");
+                    return true;
                 }
             }
             catch
             {
                 // Object is invalid in memory (e.g., player died, entity despawned)
                 await CancelCast("Target is no Longer Valid");
+                return true;
             }
 
             if (await GambitLogic.InterruptCast())
@@ -110,25 +112,26 @@ namespace Magitek.Utilities
                 return true;
             }
 
-            // The validity checks above go stale before the reads below run: CancelCast and the
-            // gambit check both await across frames, and the untargetable branch above cancels
-            // without returning — so these checks can run against a target that despawned and
-            // was freed mid-pulse. A freed target keeps a non-null reference: the null checks
-            // pass and the first field read through the stale pointer throws, killing the whole
-            // combat pulse (seen live: "Target is no Longer Targetable", then a crash in the
-            // job interrupt checks 59ms later). Skip the reads once the cast is no longer
-            // tracked, re-check validity, and treat an unreadable target like an invalid one.
+            // The validity checks above go stale before the reads below run: the gambit check
+            // yields across frames, so the target can despawn between those checks and these
+            // reads. A freed target keeps a non-null reference — the null checks pass and the
+            // first field read through the stale pointer throws, killing the whole combat
+            // pulse (seen live: a nuke target died mid-cast between the checks and the job
+            // interrupt switch). Skip the reads once the cast is no longer tracked, then keep
+            // everything that touches the target inside the try: even the IsValid re-check
+            // reads a liveness stamp through the object's pointer and can throw once the
+            // target's memory is unmapped. An unreadable target is treated like an invalid one.
             if (!CastingTime.IsRunning)
                 return true;
 
-            if (SpellTarget == null || !SpellTarget.IsValid)
-            {
-                await CancelCast("Target is no Longer Valid");
-                return true;
-            }
-
             try
             {
+                if (SpellTarget == null || !SpellTarget.IsValid)
+                {
+                    await CancelCast("Target vanished before the interrupt checks");
+                    return true;
+                }
+
                 // A revive (Phoenix Down) targets a body that is dead by definition, and the job checks
                 // below cancel any cast on a dead target unless it is that job's own raise spell — so every
                 // one of them would kill the revive on its first tracked pulse. Exempt the revive here once
@@ -222,10 +225,13 @@ namespace Magitek.Utilities
                         }
                 }
             }
-            catch
+            catch (Exception ex)
             {
-                // Object is invalid in memory (e.g., player died, entity despawned)
-                await CancelCast("Target is no Longer Valid");
+                // Object is invalid in memory (e.g., player died, entity despawned) — but any
+                // exception thrown by a job's interrupt checks lands here too, so put the real
+                // reason in the log instead of silently blaming the target.
+                Logger.WriteInfo($"[Casting] Interrupt checks failed, cancelling cast: {ex.Message}");
+                await CancelCast("Target is no Longer Readable");
             }
 
             #endregion
@@ -242,13 +248,17 @@ namespace Magitek.Utilities
 
                 if (msg != null)
                     Logger.Error(msg);
-
-                CastingTime.Stop();
-                CastingRevive = false;
             }
             catch (Exception)
             {
                 //Ignore on Purpose
+            }
+            finally
+            {
+                // The IsRunning guard in TrackSpellCast depends on this timer stopping even
+                // when StopCasting or the wait throws mid-transition.
+                CastingTime.Stop();
+                CastingRevive = false;
             }
         }
 
--- a/Magitek/Utilities/Casting.cs
+++ b/Magitek/Utilities/Casting.cs
@@ -110,97 +110,122 @@ namespace Magitek.Utilities
                 return true;
             }
 
-            // A revive (Phoenix Down) targets a body that is dead by definition, and the job checks
-            // below cancel any cast on a dead target unless it is that job's own raise spell — so every
-            // one of them would kill the revive on its first tracked pulse. Exempt the revive here once
-            // rather than carving an exception into all seven jobs, but keep the ONE cancel that is
-            // meaningful for it, the same one every healer applies to its own raise: someone else's
-            // raise landed first, so finishing the cast would spend a Phoenix Down on a claimed corpse.
-            // The target-validity checks above still run: a despawned corpse still cancels.
-            if (CastingRevive)
-            {
-                if (SpellTarget is Character corpse
-                    && (corpse.CurrentHealth > 0 || corpse.HasAura(Auras.Raise)))
-                    await CancelCast("Revive target was already raised");
+            // The validity checks above go stale before the reads below run: CancelCast and the
+            // gambit check both await across frames, and the untargetable branch above cancels
+            // without returning — so these checks can run against a target that despawned and
+            // was freed mid-pulse. A freed target keeps a non-null reference: the null checks
+            // pass and the first field read through the stale pointer throws, killing the whole
+            // combat pulse (seen live: "Target is no Longer Targetable", then a crash in the
+            // job interrupt checks 59ms later). Skip the reads once the cast is no longer
+            // tracked, re-check validity, and treat an unreadable target like an invalid one.
+            if (!CastingTime.IsRunning)
+                return true;
 
+            if (SpellTarget == null || !SpellTarget.IsValid)
+            {
+                await CancelCast("Target is no Longer Valid");
                 return true;
             }
 
-            // ReSharper disable once SwitchStatementMissingSomeCases
-            switch (RotationManager.CurrentRotation)
+            try
             {
-                case ClassJobType.BlueMage:
-                    {
-                        if (BlueMage.NeedToInterruptCast())
+                // A revive (Phoenix Down) targets a body that is dead by definition, and the job checks
+                // below cancel any cast on a dead target unless it is that job's own raise spell — so every
+                // one of them would kill the revive on its first tracked pulse. Exempt the revive here once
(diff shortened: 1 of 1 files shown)


Magitek/Utilities/Casting.cs
```diff
@@ -110,97 +110,122 @@ public static async Task<bool> TrackSpellCast()
...
+            // without returning — so these checks can run against a target that despawned and
+            // was freed mid-pulse. A freed target keeps a non-null reference: the null checks
+            // pass and the first field read through the stale pointer throws, killing the whole
+            // combat pulse (seen live: "Target is no Longer Targetable", then a crash in the
+            // job interrupt checks 59ms later). Skip the reads once the cast is no longer
+            // tracked, re-check validity, and treat an unreadable target like an invalid one.
+            if (!CastingTime.IsRunning)
+                return true;
 
+            if (SpellTarget == null || !SpellTarget.IsValid)
```
cheesegoldfish: This read is outside the try below, and `IsValid` can throw here — see the review comment for the decompiled body; it ends in `Core.Memory.Read<uint>(base.Pointer + ...)`.

When the target's page is unmapped rather than its pointer zeroed, this line throws `ReadWriteMemoryException` straight out of `TrackSpellCast` and kills the pulse — the same failure this PR fixes. It's also why the byte-identical check on line 90 is wrapped.

Simplest fix: move this block inside the `try` on line 130. The existing `catch` already does the right thing for it.

Magitek/Utilities/Casting.cs
```diff
@@ -110,97 +110,122 @@ public static async Task<bool> TrackSpellCast()
...
-                    await CancelCast("Revive target was already raised");
+            // The validity checks above go stale before the reads below run: CancelCast and the
+            // gambit check both await across frames, and the untargetable branch above cancels
+            // without returning — so these checks can run against a target that despawned and
+            // was freed mid-pulse. A freed target keeps a non-null reference: the null checks
+            // pass and the first field read through the stale pointer throws, killing the whole
+            // combat pulse (seen live: "Target is no Longer Targetable", then a crash in the
+            // job interrupt checks 59ms later). Skip the reads once the cast is no longer
+            // tracked, re-check validity, and treat an unreadable target like an invalid one.
+            if (!CastingTime.IsRunning)
```
cheesegoldfish, line 131: Two things on this guard.

**1.** It assumes `CancelCast` always reaches `CastingTime.Stop()`, but `CancelCast` wraps its whole body in `try { ... } catch (Exception) { }`. If `ActionManager.StopCasting()` or the `Coroutine.Wait` predicate throws — zone or duty transition mid-cast — the timer is never stopped, this guard sees `IsRunning == true`, and execution falls into the switch anyway. Moving `CastingTime.Stop(); CastingRevive = false;` into a `finally` in `CancelCast` makes it hold.

**2.** The body names the root cause as "the untargetable branch cancels the cast but does not return", but the `return` is never added there. Worth adding `return true;` after the `CancelCast` on line 98 too: as it stands an already-cancelled cast still runs `GambitLogic.InterruptCast()` on line 107, and on a match that logs "Canceling Cast To Execute Gambit", resets `InterruptGambitQueue`, and calls `CancelCast` a second time — bookkeeping spent on a cast that no longer exists.

Magitek/Utilities/Casting.cs
```diff
@@ -110,97 +110,122 @@ public static async Task<bool> TrackSpellCast()
...
+                            {
+                                await CancelCast();
+                            }
+                            break;
                         }
-                        break;
-                    }
+                }
+            }
+            catch
```
cheesegoldfish: Please capture and log the exception — `catch (Exception ex)` plus `Logger.WriteInfo($"...{ex.Message}")`, matching e.g. `Utilities/PhantomJobSwitcher.cs:385`.

As written this swallows every exception from nine `NeedToInterruptCast()` implementations, not just memory reads, and reports all of them as "Target is no Longer Valid". Concrete case — `Astrologian.NeedToInterruptCast()` does:

```csharp
PartyManager.VisibleMembers.Select(r => r.BattleCharacter).Count(r => r.CurrentHealth > 0 && ...)
```

A null `BattleCharacter` throws NRE. Today that surfaces in the log. With this catch, Helios is cancelled and the log says the target is invalid — on a target that's fine. The condition is deterministic, so every Helios attempt cancels, with nothing in the log to trace it. WHM has the same shape via the cached `Group.CastableAlliesWithin30`.

Narrowing the catch by type isn't an option — `ReadWriteMemoryException` and `GreyMagicException` are both `internal` to GreyMagic. Logging is the fix.

Magitek/Utilities/Casting.cs
```diff
@@ -110,97 +110,122 @@ public static async Task<bool> TrackSpellCast()
...
+            // pass and the first field read through the stale pointer throws, killing the whole
+            // combat pulse (seen live: "Target is no Longer Targetable", then a crash in the
+            // job interrupt checks 59ms later). Skip the reads once the cast is no longer
+            // tracked, re-check validity, and treat an unreadable target like an invalid one.
+            if (!CastingTime.IsRunning)
+                return true;
 
+            if (SpellTarget == null || !SpellTarget.IsValid)
+            {
+                await CancelCast("Target is no Longer Valid");
```
cheesegoldfish: Minor: this is now the third site emitting the identical `"Target is no Longer Valid"` string (also lines 92 and 228). The diagnosis in your PR body came from reading log ordering — worth keeping that possible for whoever debugs this next by giving each site its own message.

cheesegoldfish (changes requested):
Thanks for the writeup — the root-cause diagnosis is right, and the ordering rationale holds: `GambitLogic.InterruptCast()` does `await Coroutine.Yield()`, so the check at the top of the method really is stale by the time the switch runs.

One claim in the body doesn't hold, though, and it's the one that makes the new re-check look safe:

> `IsValid` is a liveness check against the object manager and does not read through the object's pointer

Decompiled from `RebornBuddy.dll`:

```csharp
public override bool IsValid {
    get {
        if (base.Pointer == IntPtr.Zero) return false;
        if (m_ConvertibleSynchronizerLength != AlignIterableStack()) return false;
        return true; } }

private uint AlignIterableStack() {
    uint num = Core.Memory.Read<uint>(base.Pointer + ...);
```

`IsValid` re-reads ObjectId at the stored address and compares it against the cached one — it does read through the pointer, and `Core.Memory.Read` throws `ReadWriteMemoryException` when the page is gone. That's exactly why the pre-existing check on line 90 sits inside a try/catch.

Two despawn shapes matter:

- **`Pointer` zeroed** — `IsValid` returns false on the first line, no memory read. Your new check handles this, and it's the case in your log: `addr: 1AC` is consistent with a read at `0 + CurrentHealth offset`, via `Casting.SpellTarget?.CurrentHealth` in `BlackMage.NeedToInterruptCast()` (`?.` tests the managed reference, which is non-null).
- **`Pointer` non-null, page unmapped** — `IsValid` throws, from outside the new try, out of `TrackSpellCast`.

So the change fixes the crash you hit but leaves open the general case it was written for. Comments inline.

Structurally, one contiguous `try` from line 88 to 229 plus a `return true;` after the cancel on line 98 would remove the duplicated guard, the duplicated catch body, and the unprotected seam at line 107 — and comes out smaller than what's here now.

Separately, and **not blocking this PR**: `CheckForSuccessfulCast` has the same crash class untouched — `await Coroutine.Wait(3000, () => auraTarget.HasAura(Aura, true) || !auraTarget.IsValid || ...)` short-circuits left to right, so `HasAura` is read before the `IsValid` guard, on a possibly-freed object, for up to 3s. Worth its own PR.

Magitek/Utilities/Casting.cs
HomoYouDidnt: Confirmed — I had read the property body but not the stamp getter it calls, which does read through `base.Pointer`. My claim in the PR body was wrong; corrected there too. The re-check now lives inside the try as its first statement, so both despawn shapes — pointer zeroed and page unmapped — end in a cancel instead of a thrown pulse.

Magitek/Utilities/Casting.cs
HomoYouDidnt: Both taken. `CastingTime.Stop()` and `CastingRevive = false` moved into a `finally` in `CancelCast`, so the IsRunning guard holds even if StopCasting or the wait throws mid-transition. Added the missing `return true;` to the untargetable branch — and to the catch beside it, which had the same cancel-then-fall-through shape (gambit bookkeeping spent on a dead cast).

Magitek/Utilities/Casting.cs
HomoYouDidnt: Done — `catch (Exception ex)` with `Logger.WriteInfo($"[Casting] Interrupt checks failed, cancelling cast: {ex.Message}")`, matching the PhantomJobSwitcher pattern, so a deterministic NRE in a job check reads as what it is instead of a phantom invalid target. Agreed on not catching by type with the GreyMagic exception types being internal.

Magitek/Utilities/Casting.cs
HomoYouDidnt: Each site has its own message now: the top check keeps "Target is no Longer Valid", the re-check says "Target vanished before the interrupt checks", and the catch cancels with "Target is no Longer Readable" plus the logged exception.

Magitek/Utilities/Casting.cs
```diff
@@ -110,97 +112,126 @@ public static async Task<bool> TrackSpellCast()
...
-            switch (RotationManager.CurrentRotation)
+            try
             {
-                case ClassJobType.BlueMage:
-                    {
-                        if (BlueMage.NeedToInterruptCast())
+                if (SpellTarget == null || !SpellTarget.IsValid)
+                {
+                    await CancelCast("Target vanished before the interrupt checks");
+                    return true;
```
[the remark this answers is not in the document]
HomoYouDidnt: Confirmed and taken — a cast that completes during the gambit yields would have hit this branch with the timer still running, been "cancelled", and skipped the success bookkeeping. Added `if (!Core.Me.IsCasting) return false;` after the gambit block, mirroring the entry check, so a finished cast reaches `CheckForSuccessfulCast` and lands in `LastSpell`/`SpellCastHistory`; the invalid-target re-check now only ever sees a cast that is still live.

cheesegoldfish (approved):

PR #234, opened 2026-08-17
FightLogic - Stop the Doom heal re-firing every pulse while moving

HomoYouDidnt:
## What this changes

The heal-to-full Doom response re-fires its heal every pulse while the healer is moving. The heal has a cast bar, so each attempt fails instantly mid-movement, the Doom is still present next pulse, and the response fires again — observed live at ~25 attempts per second, sustained for the full duration of the movement.

The fix: skip the response while moving when the heal has a cast time, unless Swiftcast is up (instant heals still go out on the move). The heal then fires on the first stationary pulse — which is also what actually happened in the incident once movement stopped.

## Tested

Live Forked Tower: Blood run, both Scholar and Sage. Before the guard: one Doom carried while moving produced a sustained ~25/s re-fire storm in the log. After: dozens of Doom cycles across both jobs, every one answered with a single cast line and cleansed.

## Sources

Live log evidence as above; the failure and the fix were both observed in the same session.

## Removals

None — one guard added, matching the cast-while-moving pattern used elsewhere in the codebase.

### DIFF
--- a/Magitek/Logic/Roles/CommonFightLogic.cs
+++ b/Magitek/Logic/Roles/CommonFightLogic.cs
@@ -192,8 +192,14 @@ namespace Magitek.Logic.Roles
             // A heal with a cast bar fails instantly while moving, and this check runs ahead of
             // discretionary healing on every pulse — without this guard a Doom carried while
             // moving re-fires the failed cast every pulse (~25/s, seen live) until movement
-            // stops. Swiftcast makes the cast instant, so it is exempt.
-            if (ff14bot.Managers.MovementManager.IsMoving && heal.AdjustedCastTime > System.TimeSpan.Zero && !Core.Me.HasAura(Utilities.Auras.Swiftcast))
+            // stops. Swiftcast and Dualcast both make the cast instant, so both are exempt —
+            // the same pair the shared cast gate whitelists. Dualcast is not RedMage-only:
+            // Occult Crescent's Phantom Red Mage grants the same status on any job, so a
+            // Scholar in the field ops carries it through most of a fight (observed in game).
+            if (ff14bot.Managers.MovementManager.IsMoving
+                && heal.AdjustedCastTime > System.TimeSpan.Zero
+                && !Core.Me.HasAura(Utilities.Auras.Swiftcast)
+                && !Core.Me.HasAura(Utilities.Auras.Dualcast))
                 return false;
 
             var doomed = FightLogic.DoomedHealTarget();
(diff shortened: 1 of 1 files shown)


HomoYouDidnt:
Follow-up pushed (326bb7d9): the moving guard now exempts Dualcast as well as Swiftcast.

Found in game tonight rather than in review. Occult Crescent's Phantom Red Mage grants **Dualcast (status 1249 — the same status the Red Mage job uses)** to whatever job you are actually playing, so a Scholar running that phantom job in the field ops has it up through most of a fight: sampled live it was gained and consumed on a roughly 1.3s cycle for minutes at a time. With the guard whitelisting only Swiftcast, every one of those windows made this return false while moving, skipping a Doom heal that would in fact have been instant — and Doom is common in exactly that content (field CEs, the Necromancer's self-Doom).

The pair now matches what the shared cast gate in `SpellDataExtensions.Check` treats as instant, so the early bail and the real gate agree. Worth noting for anyone reading the guard: Dualcast is not Red Mage-only in practice.

HomoYouDidnt:
Correction to my own comment above, now that I have measured it rather than reasoned about it.

I claimed the Swiftcast-only guard was skipping Doom heals during Dualcast windows. That is **wrong**, and the Dualcast exemption I pushed is behaviourally redundant.

I instrumented the Scholar decline path to print `dualcast=` alongside the live value of `Spells.Succor.AdjustedCastTime`, and ran it through several hours of Occult Crescent and alliance content while on Phantom Red Mage. The correlation is exact, with no exceptions:

| dualcast | AdjustedCastTime | samples |
|---|---|---|
| False | 1984 ms | 12 |
| True | **0 ms** | 2 |

So `AdjustedCastTime` **does** reflect Dualcast. That means the guard's `heal.AdjustedCastTime > TimeSpan.Zero` term was already false during exactly those windows, the heal was already permitted, and my added `!Core.Me.HasAura(Auras.Dualcast)` term cannot change the outcome in any state. It is harmless, and it arguably documents intent for the next reader, but it fixed nothing and I should not have described it as fixing something.

Two things are still worth keeping from it.

`AGENTS.md` says in the SpellData timing section that AdjustedCastTime *"does not account for Swiftcast or other instant-cast buffs"*. That appears to be inaccurate, at least for Dualcast — and it is the kind of sentence someone writes a movement gate against, adding aura checks they do not need, or worse, trusting a cast time that is not real. I am happy to send a doc correction if you read the measurement the same way.

And the underlying game fact holds regardless: Occult Crescent's Phantom Red Mage grants Dualcast — status **1249**, the same id the Red Mage job uses, not a phantom-specific one — to whatever job you are actually playing. On a Scholar in field ops it cycles roughly every 1.3 seconds for most of a fight, which is why it turned up in the logs at all.

Apologies for the noise on a merged PR. I would rather correct the record than leave a wrong rationale sitting in the history where someone builds on it later.

PR #243, opened 2026-08-18
SCH - Fix Aetherpact never releasing the tank once they are back to full health

HomoYouDidnt:
## What this changes

Aetherpact currently starts correctly and then never releases. Two conditions in `CanDeAetherpact` prevent it, and either one alone is enough to make the break unreachable.

**The Fey Union check could never be satisfied.** It read:

```csharp
if (!unit.HasAura(Auras.FeyUnion) || !unit.HasAura(Auras.FeyUnion2))
    return false;
```

which rejects the unit unless it holds **both** ids at once. Fey Union applies as one of two ids — 1222 and 1223, both named "Fey Union" in the status sheet — never both together, so this always returned false and the method could never produce a target. The three other Fey Union checks in the same file already treat them as alternatives; this is the only one that negates them. Now rejects only a unit carrying neither.

**The health comparison was inverted relative to its own setting.** It read:

```csharp
if (unit.CurrentHealthPercent >= ScholarSettings.Instance.BreakAetherpactHp)
    return false;
```

`BreakAetherpactHp` defaults to 100 and the option is presented as *"Break Aetherpact If Tank Is Full HP Only With N enemies"*. So the code declined to break precisely when the tank was full — the case the setting exists to describe — and would only have released while they were still hurt. Flipped to `<`, so a topped-up tank ends the channel.

**One guard added, because the fix makes a latent hazard reachable.** Aetherpact is a toggle: cast at someone who already has Fey Union it ends the channel, cast again it starts a new one. The aura does not clear the instant the break lands, so a second pulse could re-establish the pact it had just released. The start path already guards on `Casting.LastSpell == Spells.Aetherpact`; the break path now does too. This only matters now that the break can fire at all.

The enemy-count condition above these is unchanged and was already correct — it holds the pact while more than N enemies are on the tank.

## Why

Reported from play: the fairy takes the pact at the configured health threshold and then channels indefinitely, draining Fairy Gauge on a tank who is back at full. With both conditions as they were, that is the only possible behaviour.

## Verified against game data

- Status ids 1222 and 1223 both resolve to "Fey Union" in the client's own status sheet, confirming they are alternative applications of one effect rather than two co-applied auras.
- The option's own label and its 100% default establish the intended direction of the health comparison.

None. Both changes are read off the aura data and the setting's stated meaning.

## Anything removed

No conditions removed. Two comparisons corrected, one guard added.

## In-game test plan

1. Scholar with a fairy out, in a party, Aetherpact enabled. Let a tank drop below the start threshold (75% by default) and confirm Fey Union begins as it does today.
2. Heal the tank to full with three or fewer enemies on them. Expect the pact to release at 100%, and Fairy Gauge to stop draining.
3. Watch the pulse after the release: the pact must not immediately restart. If it does, the toggle guard is not holding.
4. Repeat with more than N enemies still on the tank and confirm the pact is *held* rather than released, which is the existing enemy-count behaviour.

## Build

- [x] `dotnet build` clean in `Magitek/`
- [x] No unrelated files in the diff

### DIFF
--- a/Magitek/Logic/Scholar/Buff.cs
+++ b/Magitek/Logic/Scholar/Buff.cs
@@ -489,6 +489,15 @@ namespace Magitek.Logic.Scholar
                 if (unit.CurrentHealthPercent < ScholarSettings.Instance.BreakAetherpactHp)
                     return false;
 
+                // Both thresholds accept 1-100 independently, so a release threshold at or below the
+                // engage threshold would release a tank who immediately qualifies to be re-pacted, and
+                // the pair would alternate until the tank climbed past the engage value - burning
+                // gauge and an oGCD slot each cycle. Requiring the tank to be above the ENGAGE
+                // threshold too makes the two settings coherent whatever they are set to, without
+                // rejecting or silently rewriting the user's numbers.
+                if (unit.CurrentHealthPercent < ScholarSettings.Instance.AetherpactHealthPercent)
+                    return false;
+
                 // Fey Union applies as one of two ids (1222 / 1223), never both at once, so
                 // requiring both here could never be satisfied and this method could never return
                 // a target. Reject only a unit carrying neither. The three other Fey Union checks
--- a/Magitek/Logic/Scholar/Buff.cs
+++ b/Magitek/Logic/Scholar/Buff.cs
@@ -19,6 +19,16 @@ namespace Magitek.Logic.Scholar
         // Prevents double summoning of fairy
         public static DateTime FairySummonCooldown = DateTime.Now;
 
+        // Aetherpact is a toggle, so a second cast inside the latency window after one lands can
+        // re-establish the pact that was just released. Debounced on time rather than on
+        // Casting.LastSpell: with damage disabled, or no attack target, the Scholar can go a long
+        // while casting nothing else, and LastSpell would then sit on Aetherpact indefinitely and
+        // latch both paths off — the break permanently, which is the failure this file is fixing.
+        private static DateTime AetherpactToggle = DateTime.MinValue;
+
+        private static bool AetherpactToggledRecently =>
+            (DateTime.Now - AetherpactToggle).TotalMilliseconds < 1500;
+
         public static async Task<bool> SummonPet()
         {
             if (Core.Me.Pet != null)
@@ -375,7 +385,7 @@ namespace Magitek.Logic.Scholar
             if (!Globals.PartyInCombat)
                 return false;
 
-            if (Casting.LastSpell == Spells.Aetherpact)
+            if (AetherpactToggledRecently)
                 return false;
 
             if (!ActionManager.HasSpell(Spells.Aetherpact.Id))
@@ -399,7 +409,11 @@ namespace Magitek.Logic.Scholar
             //if (Casting.LastSpell != Spells.Biolysis || Casting.LastSpell != Spells.ArtOfWar || Casting.LastSpell != Spells.Adloquium || Casting.LastSpell != Spells.Succor)
             //    if (await Spells.Ruin2.Cast(Core.Me.CurrentTarget))
             //        return true;
-            return await Spells.Aetherpact.Cast(aetherpactTarget);
+            if (!await Spells.Aetherpact.Cast(aetherpactTarget))
+                return false;
+
+            AetherpactToggle = DateTime.Now;
+            return true;
 
             bool CanAetherpact(GameObject unit)
             {
@@ -429,10 +443,9 @@ namespace Magitek.Logic.Scholar
                 return false;
 
             // Aetherpact is a toggle: cast at a unit that already has Fey Union it ends the channel,
-            // cast again it starts a new one. The aura does not clear the instant the break lands, so
-            // without this the next pulse can re-establish the pact it just released. The same guard
-            // sits on the start path; it only matters here now that this method can actually fire.
-            if (Casting.LastSpell == Spells.Aetherpact)
+            // cast again it starts a new one. The aura does not clear the instant the break lands,
+            // so without a debounce the next pulse can re-establish the pact it just released.
+            if (AetherpactToggledRecently)
                 return false;
 
             if (!Group.CastableAlliesWithin30.Any(r => r.HasAura(Auras.FeyUnion) || r.HasAura(Auras.FeyUnion2)))
@@ -443,7 +456,11 @@ namespace Magitek.Logic.Scholar
             if (aetherpactTarget == null)
                 return false;
 
-            return await Spells.Aetherpact.Cast(aetherpactTarget);
+            if (!await Spells.Aetherpact.Cast(aetherpactTarget))
+                return false;
+
+            AetherpactToggle = DateTime.Now;
+            return true;
 
             bool CanDeAetherpact(GameObject unit)
             {
--- a/Magitek/Logic/Scholar/Buff.cs
+++ b/Magitek/Logic/Scholar/Buff.cs
@@ -464,6 +464,14 @@ namespace Magitek.Logic.Scholar
 
             bool CanDeAetherpact(GameObject unit)
             {
+                // The pact is only ever placed on a tank (see CanAetherpact), so the unit whose health
+                // and surroundings decide whether to release it has to be that tank. Excluding
+                // ourselves matters because the two Fey Union ids may not both sit on the recipient:
+                // if one of them lands on the Scholar, we would otherwise weigh OUR health and OUR
+                // nearby enemies and release a pact on a tank who is still hurt.
+                if (unit == null || unit == Core.Me || !unit.IsTank())
+                    return false;
+
                 if (unit.EnemiesNearby(6).Count() > ScholarSettings.Instance.AetherpactEnemies)
                     return false;
 
--- a/Magitek/Logic/Scholar/Buff.cs
+++ b/Magitek/Logic/Scholar/Buff.cs
@@ -428,6 +428,13 @@ namespace Magitek.Logic.Scholar
             if (!ActionManager.HasSpell(Spells.Aetherpact.Id))
                 return false;
 
+            // Aetherpact is a toggle: cast at a unit that already has Fey Union it ends the channel,
+            // cast again it starts a new one. The aura does not clear the instant the break lands, so
+            // without this the next pulse can re-establish the pact it just released. The same guard
+            // sits on the start path; it only matters here now that this method can actually fire.
+            if (Casting.LastSpell == Spells.Aetherpact)
+                return false;
+
             if (!Group.CastableAlliesWithin30.Any(r => r.HasAura(Auras.FeyUnion) || r.HasAura(Auras.FeyUnion2)))
                 return false;
 
@@ -443,10 +450,19 @@ namespace Magitek.Logic.Scholar
                 if (unit.EnemiesNearby(6).Count() > ScholarSettings.Instance.AetherpactEnemies)
                     return false;
 
-                if (unit.CurrentHealthPercent >= ScholarSettings.Instance.BreakAetherpactHp)
+                // Break once the tank is topped up, which is what the option says on the tin
+                // ("Break Aetherpact If Tank Is Full HP Only With N enemies") and what the 100%
+                // default describes. The comparison was the other way round, so the pact was held
+                // exactly while the tank no longer needed it and released only while they were
+                // still hurt — the opposite of the setting, and a straight waste of Fairy Gauge.
+                if (unit.CurrentHealthPercent < ScholarSettings.Instance.BreakAetherpactHp)
                     return false;
 
-                if (!unit.HasAura(Auras.FeyUnion) || !unit.HasAura(Auras.FeyUnion2))
+                // Fey Union applies as one of two ids (1222 / 1223), never both at once, so
+                // requiring both here could never be satisfied and this method could never return
+                // a target. Reject only a unit carrying neither. The three other Fey Union checks
+                // in this file already test them as alternatives.
+                if (!unit.HasAura(Auras.FeyUnion) && !unit.HasAura(Auras.FeyUnion2))
                     return false;
 
                 return true;
--- a/Magitek/Logic/Scholar/Buff.cs
+++ b/Magitek/Logic/Scholar/Buff.cs
@@ -464,6 +464,12 @@ namespace Magitek.Logic.Scholar
 
             bool CanDeAetherpact(GameObject unit)
             {
+                // Releasing at all is opt-out: some Scholars run the pact as a sustained regen on the
+                // tank and would rather it never drop, which the HP threshold alone cannot express -
+                // there is no value of it that means "never".
+                if (!ScholarSettings.Instance.BreakAetherpact)
+                    return false;
+
                 // The pact is only ever placed on a tank (see CanAetherpact), so the unit whose health
                 // and surroundings decide whether to release it has to be that tank. Excluding
                 // ourselves matters because the two Fey Union ids may not both sit on the recipient:
--- a/Magitek/Models/Scholar/ScholarSettings.cs
+++ b/Magitek/Models/Scholar/ScholarSettings.cs
@@ -525,6 +525,10 @@ namespace Magitek.Models.Scholar
         public float AetherpactHealthPercent { get; set; }
 
         [Setting]
+        [DefaultValue(true)]
+        public bool BreakAetherpact { get; set; }
+
+        [Setting]
         [DefaultValue(100.0f)]
         public float BreakAetherpactHp { get; set; }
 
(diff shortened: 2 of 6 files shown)


Magitek/Logic/Scholar/Buff.cs
```diff
@@ -428,6 +428,13 @@ public static async Task<bool> BreakAetherpact()
             if (!ActionManager.HasSpell(Spells.Aetherpact.Id))
                 return false;
 
+            // Aetherpact is a toggle: cast at a unit that already has Fey Union it ends the channel,
+            // cast again it starts a new one. The aura does not clear the instant the break lands, so
+            // without this the next pulse can re-establish the pact it just released. The same guard
+            // sits on the start path; it only matters here now that this method can actually fire.
+            if (Casting.LastSpell == Spells.Aetherpact)
+                return false;
```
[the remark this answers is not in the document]
HomoYouDidnt: Confirmed and fixed in dd234d5b — and this was a defect in the guard I added, not in the original code, which makes it the worst kind: a fix that reintroduces the bug it was fixing.

You are right about the mechanism. `Casting.LastSpell` only changes when something else is successfully cast. A Scholar with damage disabled, or with no attack target, can go a long time casting nothing at all — and the Aetherpact channel itself does not update it. So `LastSpell` sits on `Aetherpact` indefinitely and the guard latches `BreakAetherpact()` off permanently, which is precisely the "pact never releases, gauge keeps draining" failure this PR exists to fix.

Replaced with a time debounce, using the pattern already in this file for `FairySummonCooldown`:

```csharp
private static DateTime AetherpactToggle = DateTime.MinValue;

private static bool AetherpactToggledRecently =>
    (DateTime.Now - AetherpactToggle).TotalMilliseconds < 1500;
```

stamped at both cast sites and checked on both paths. 1500ms covers the animation lock plus the aura settling after the toggle, and it expires on its own rather than waiting on an unrelated event.

I applied it to the **start** path as well, replacing the original `Casting.LastSpell` guard there. That guard has the same latch, and although its consequence is milder — refusing to re-establish a pact rather than refusing to release one — it is the same defect, and leaving the two paths guarding the same toggle by different rules seemed worse than changing one line of existing code. Flagging it explicitly since it is a pre-existing guard rather than something this PR introduced.

Magitek/Logic/Scholar/Buff.cs
```diff
@@ -436,17 +456,30 @@ public static async Task<bool> BreakAetherpact()
...
+                // still hurt — the opposite of the setting, and a straight waste of Fairy Gauge.
+                if (unit.CurrentHealthPercent < ScholarSettings.Instance.BreakAetherpactHp)
                     return false;
 
-                if (!unit.HasAura(Auras.FeyUnion) || !unit.HasAura(Auras.FeyUnion2))
+                // Fey Union applies as one of two ids (1222 / 1223), never both at once, so
+                // requiring both here could never be satisfied and this method could never return
+                // a target. Reject only a unit carrying neither. The three other Fey Union checks
+                // in this file already test them as alternatives.
+                if (!unit.HasAura(Auras.FeyUnion) && !unit.HasAura(Auras.FeyUnion2))
```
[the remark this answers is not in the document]
HomoYouDidnt: Fixed in f175cbeb, though I want to separate the two halves of this because one of them does not hold up.

**The cited evidence is not there.** I checked `Magitek/Resources/StatusList.json`, which the comment names directly. Both entries are plain, with no recipient-side labelling:

```
1222  Fey Union
1223  Fey Union
```

The live client's status sheet agrees — I looked both ids up through the running game and each returns `Fey Union`, same name, nothing distinguishing them. So "the bundled StatusList.json even labels 1223 as the 'received' variant" is not accurate, and I would not want that repeated as established fact later.

**The concern behind it is still worth acting on.** I cannot currently prove which actor each id lands on. If they are two applications on the recipient, my `&&` was fine; if one of them sits on the Scholar, then a Scholar sitting at the break threshold could satisfy the predicate, and the health and enemy-count checks would then be measured against *me* rather than the bonded tank — releasing a pact on a tank who is still hurt. That failure is bad enough, and cheap enough to exclude, that it does not need the id question settled first.

So the fix is deliberately assumption-free about the ids:

```csharp
if (unit == null || unit == Core.Me || !unit.IsTank())
    return false;
```

The pact is only ever established on a tank — `CanAetherpact` requires `IsTank()` — so the unit whose health decides the release has to be that same tank. This is correct whether the ids are split across actors or not, and it costs nothing if they are not.

One consequence worth stating rather than hiding: a pact the player established manually on a non-tank will not be auto-released by this path. Given the routine only ever creates them on tanks, that seemed the right trade against the alternative of weighing the wrong actor's health.

If you happen to know which of 1222/1223 is the recipient-side status, I would still like to record it — I can settle it from a live pact next time I have one up, but I would rather not assert it either way in the meantime.

HomoYouDidnt:
Validated in a live Windurst: The Third Walk alliance raid, and the Fey Union id question I left open above is now settled from the combat log rather than guessed.

**The ids are split across actors, but not the way the review suggested.** From ACT status lines during a real pact:

```
26| 14:52:23.088  id=0x4C6 (1222)  Fey Union  dur=9999  src=Eos  tgt=Eos           <- the FAIRY
26| 14:52:23.088  id=0x4C7 (1223)  Fey Union  dur=60    src=Eos  tgt=Jesse James   <- the RECIPIENT
30| 14:52:28.968  both removed
```

So 1222 is the fairy's own status and 1223 is the recipient's. Neither lands on the Scholar — I also checked 300 seconds of 250ms aura sampling across a pact window and the Scholar never carries either. The fairy is a pet and so is never in `Group.CastableAlliesWithin30`, which means the specific failure described earlier (the Scholar selecting itself as a break candidate) could not occur. The `unit != Core.Me` guard is therefore belt-and-braces; I have left it in because it costs nothing, and the `IsTank()` half still usefully constrains the pick.

**The fix works.** Four pacts in the raid, all released:

| established | released | held | on |
|---|---|---|---|
| 14:48:21.874 | 14:48:26.422 | 4.5s | Jesse James (GNB) |
| 14:51:40.791 | 14:51:44.225 | 3.4s | Mala Ether |
| 14:51:52.554 | 14:51:54.428 | 1.9s | Mala Ether |
| 14:52:23.088 | 14:52:28.968 | 5.9s | Jesse James (GNB) |

The HP column on the last one shows why it released: the tank was at 309,266 when the pact began and reached 340,192 — full — at 14:52:25, with the release at 14:52:28.968. Channel while hurt, release on full. Before this change it would have kept channelling and draining Fairy Gauge on a tank at 100%.

Releases landed 1.5–3.9s after full HP, comfortably outside the 1500ms debounce, and nothing re-pacted afterwards — so the toggle guard prevents the re-establish race without blocking legitimate releases.

HomoYouDidnt:
Correcting two errors in my validation comment above. The conclusion is unchanged — the fix works — but the evidence I attached was wrong in two ways, and the corrected version is actually stronger.

**1. Two of the four pacts in that table were not this routine.** They were a second Scholar in the alliance self-pacting. The fairies are distinguishable by actor id, and the dispatch lines confirm the caster:

```
14:51:39.991  Mala Ether -> Mala Ether      fairy 40003F2C   <- NOT us
14:48:21.117  Kairos Himeros -> Jesse James fairy 40003FC0   <- us
```

**2. The HP numbers were max HP, not current HP.** On ACT status lines (type 26/30) field 11 is the *target's max HP*; current HP appears only on ability lines (type 21/22). So "309,266 -> 340,192 (full)" was a max-HP buff landing mid-fight, not the tank being healed. That sentence should be struck.

**Corrected evidence.** Rather than inferring from HP sampling, the routine logs the break explicitly, so the pairing is direct — six establishes, six breaks, no unmatched pact:

| # | establish | break | held |
|---|---|---|---|
| 1 | 18:48:22.708 | 18:48:27.858 | 5.2s |
| 2 | 18:52:23.572 | 18:52:30.444 | 6.9s |
| 3 | 18:56:27.474 | 18:56:38.846 | 11.4s |
| 4 | 18:56:46.775 | 18:56:52.920 | 6.1s |
| 5 | 18:57:31.061 | 18:57:46.459 | 15.4s |
| 6 | 18:58:01.206 | 18:58:04.343 | 3.1s |

Each break is a distinct dispatch, not the establish line repeated. Cross-checked against the combat log, and correcting for the ~1.6s clock skew between the two sources (measured off the establish dispatches), every Fey Union removal follows its break by 0.10-0.16s — a consistent server round-trip. That rules out the alternative reading I could not exclude from HP data alone: that the pacts were simply expiring on fairy gauge with the break incidental.

The break threshold was at its default of 100 for this run, so all six released on a full-health tank. No re-pact occurred inside the 1500ms debounce; the shortest gap between a break and the next establish was 8.5s, each time because the tank had taken fresh damage.

The Fey Union id finding in my previous comment (1222 on the fairy, 1223 on the recipient) is unaffected — those came from the status lines' id and target fields, which I read correctly.

Magitek/Logic/Scholar/Buff.cs
```diff
@@ -19,6 +19,16 @@ internal static class Buff
...
 
+        // Aetherpact is a toggle, so a second cast inside the latency window after one lands can
+        // re-establish the pact that was just released. Debounced on time rather than on
+        // Casting.LastSpell: with damage disabled, or no attack target, the Scholar can go a long
+        // while casting nothing else, and LastSpell would then sit on Aetherpact indefinitely and
+        // latch both paths off — the break permanently, which is the failure this file is fixing.
+        private static DateTime AetherpactToggle = DateTime.MinValue;
+
+        private static bool AetherpactToggledRecently =>
+            (DateTime.Now - AetherpactToggle).TotalMilliseconds < 1500;
```
cheesegoldfish, line 30: Aetherpact has a 3 second recast, this doesn't do anything. 

I suspect we should at least have a minimum amount of time aetherpact should be enabled on the tank, if it engages, it should engage for at least 6 seconds or so. 

I use aetherpact all the time to keep a tank at full during a large pull, and so even if the tank is full hp, I personally don't want aetherpact dropping because it acts as the regen. 

So unrelated to this, Aetherpact disengage probably needs it's own toggle if it doesn't have one.

Magitek/Logic/Scholar/Buff.cs
```diff
@@ -436,17 +456,38 @@ public static async Task<bool> BreakAetherpact()
...
                 if (unit.EnemiesNearby(6).Count() > ScholarSettings.Instance.AetherpactEnemies)
                     return false;
 
-                if (unit.CurrentHealthPercent >= ScholarSettings.Instance.BreakAetherpactHp)
+                // Break once the tank is topped up, which is what the option says on the tin
+                // ("Break Aetherpact If Tank Is Full HP Only With N enemies") and what the 100%
+                // default describes. The comparison was the other way round, so the pact was held
+                // exactly while the tank no longer needed it and released only while they were
+                // still hurt — the opposite of the setting, and a straight waste of Fairy Gauge.
+                if (unit.CurrentHealthPercent < ScholarSettings.Instance.BreakAetherpactHp)
```
cheesegoldfish, line 489: Since this is now making the Hp threashold work. Might as well add the UI setting for it, and a toggle to enable or disable breaking on HP threshold.

HomoYouDidnt:
Thanks — the regen point is the important one and I want to address that first, because I think the behaviour you are describing is already protected and predates this PR.

**Large pulls already hold the pact.** The first check in `CanDeAetherpact` is master's, untouched here:

```csharp
if (unit.EnemiesNearby(6).Count() > ScholarSettings.Instance.AetherpactEnemies)
    return false;
```

With the default of 3, the release cannot fire while more than three enemies are within 6y of the tank. So during the big pulls you are describing, Aetherpact keeps channelling as a regen exactly as you want, at full HP or not. This PR only changes what happens at three enemies or fewer — single target and small packs, where the tank is topped off and the gauge is draining for nothing.

**On the 3s recast** — you are right that one exists, but it gates re-engaging rather than disengaging, so the debounce is not inert. Measuring every toggle across a night of play:

```
BREAK -> ENGAGE   minimum 3.928s   (n=6)    <- the recast you describe
ENGAGE -> BREAK   minimum 1.504s   (n=29)   <- not gated by it
```

Twenty-nine of the engage-to-break gaps sit exactly on the 1500ms debounce, which is the only thing stopping the release from firing on the pulse straight after the engage. Without it the pact could be dropped almost immediately after being placed.

Worth being clear about what was actually broken, since it is easy to read this as a behaviour change rather than a repair. On master the release could never fire at all:

```csharp
if (unit.CurrentHealthPercent >= BreakAetherpactHp) return false;   // releases only while HURT
if (!unit.HasAura(FeyUnion) || !unit.HasAura(FeyUnion2)) return false;  // needs BOTH ids
```

The HP comparison is inverted against what the option says on the tin, and the aura check requires both Fey Union ids at once — but they never coexist on one actor. 1222 sits on the fairy and 1223 on the recipient, confirmed from the combat log during a live pact. So the second condition can never be satisfied and the release was dead code.

**Your two suggestions both sound right to me and neither exists today.**

A minimum engagement time is not implemented at all — 6s would be easy to add and would stop the short establish/release churn that shows up when a tank oscillates around the threshold.

You are also correct that there is no disengage toggle. `BreakAetherpactHp` is a threshold only, and there is no value that means "never release" — setting it to 100 still releases at full HP, which is precisely the case you do not want during a pull.

I would rather not fold either into this PR, since it is a fix and those are features, and you have asked for one mechanism per PR. Happy to put them up separately — say whether you want the minimum duration, the toggle, or both, and whether the toggle should default to the current behaviour or to never releasing.

Magitek/Logic/Scholar/Buff.cs
cheesegoldfish: Oh just to correct myself on this one. Apparently we still require < 3 enemies later, so that sill does what I want and keeps it on during aoe tanking.

HomoYouDidnt:
Done, pushed. Two settings, both in the Aetherpact panel:

- **Break Aetherpact** — new on/off toggle, defaults to on so current behaviour is unchanged. Turning it off means the pact is never released on HP at all, which is the case the threshold could not express: there was no value of it that meant "never".
- **Once The Tank Reaches [ 100 ] Health Percent** — the existing `BreakAetherpactHp`, which had no UI at all. It has been in the settings file since before this PR, so anyone who edited the JSON by hand had it working; from the UI it was invisible.

I also reworded the enemy row while I was in there. It previously read *"Break Aetherpact If Tank Is Full HP Only With [3] or less Enemies"*, which baked the 100% into the label and would have been wrong the moment the threshold became editable. It now reads as a continuation:

```
[x] Break Aetherpact
    Once The Tank Reaches [100] Health Percent
    And There Are [3] or less Enemies
```

The old string is left in the resource files rather than deleted, in case anything else references it.

Three new strings added to both `Resources.resx` and `Resources.zh-CN.resx`, with the `Resources.Designer.cs` properties written by hand since it only regenerates in Visual Studio.

And thank you for the correction on the enemy gate — you got there before I posted, but it is worth having the number on the record: with the default of 3, that check is what keeps the pact channelling through AoE tanking regardless of the tank's HP.

**Tested:** Not tested in game — the settings and logic build clean but I have not exercised the new toggle in a duty. The behaviour with it left on is the same code path this PR already covered.

HomoYouDidnt:
One of these I have fixed; the other I believe is refuted by the combat log.

**Hysteresis (P2) — fixed.** You are right that the two thresholds accept 1-100 independently and can be set to fight each other: release at or below the engage value means the tank qualifies to be re-pacted the instant it is released, and the pair alternates until they climb past the engage threshold. Rather than rejecting or rewriting the user's numbers, the release now also requires the tank to be above the *engage* threshold, which makes the two coherent at any setting:

```csharp
if (unit.CurrentHealthPercent < ScholarSettings.Instance.AetherpactHealthPercent)
    return false;
```

**Dissolve Union (P1) — the release already reaches the server as Dissolve Union.** From the combat log of a clear, filtering on action id rather than name:

```
14:48:21.117  1D0D  Aetherpact
14:48:26.288  1EBD  Dissolve Union
14:52:21.973  1D0D  Aetherpact
14:52:28.834  1EBD  Dissolve Union
14:56:25.889  1D0D  Aetherpact
14:56:37.251  1EBD  Dissolve Union
```

Every establish is `1D0D` (7437) and every release is `1EBD` (7869), perfectly interleaved. So casting `Spells.Aetherpact` at a unit that already carries Fey Union does reach the server as Dissolve Union — the masking resolves below the routine, and the Fey Union statuses are removed 0.10-0.16s after each of those release dispatches. The pact is not left active and the gauge drain does stop.

I want to flag that this one nearly caught me out in the other direction, because I had recorded exactly the opposite. Earlier in this work I searched the combat log for "Aetherpact", found dispatches only for the establishes, and concluded the release produced no line at all. It does — under its own name, which I had not searched for. Your comment is what sent me back to check by id.

Given that, I have not switched the call to `Spells.DissolveUnion` explicitly. It would be a reasonable readability change and I am happy to make it if you would prefer the intent spelled out at the call site, but on the evidence it is not a behaviour fix, and the same masked call is used by the three other Fey Union paths in this file.

PR #247, opened 2026-08-19
FightLogic - Mitigate the marked AoEs and the stack in AAC Heavyweight (M4)

HomoYouDidnt:
Lindwurm had its tank buster, The Fixer and Splattershed catalogued, but the marker mechanics were missing. All figures below are measured from a live pull as a share of each victim's max HP.

**Venomous Scourge (46208)** — marked AoE on every player. 22 hits across all eight, up to 62.2%, which is comparable to the Splattershed already listed, so it sits in the same chain.

**Grotesquerie (46209)** — inflicts Burning Grotesquerie on all players, resolving as a marked AoE when it expires. Worth being explicit about why this id: the damage lands roughly 18 seconds later under a *different* name, Dramatic Lysis, measured at 73.6% across all eight, and Dramatic Lysis has no cast line of its own. The Grotesquerie cast is the only thing fight logic can react to.

**Fourth-wall Fusion — catalogued by head marker (93), not cast id.** This is the fight's most dangerous mechanic: 317.9% of a health bar across six players, lethal if the stack is missed. It has no type-20 cast line whatsoever, so cast-id detection cannot see it at all. Its only antecedent cast is Ravenous Reach 11 to 12 seconds earlier, which is itself a positional cleave and would fire mitigation at the wrong moment.

Marker 93 is the stack marker. That is not inferred from this fight alone: it precedes all three Fourth-wall Fusion bursts here on a single target, and it also lands in the same millisecond as Mog Stone IV in Thornmarch, which is likewise a stack. Two unrelated bosses, single target both times.

I deliberately did not add marker 311 alongside it. That one landed on all eight with Mograin of Death and is the spread-on-everyone marker; the spreads in this fight are already covered by cast id above, so listing it would double-detect.

One calibration note for the reviewer: I did not promote anything to `BigAoes` here. The Fixer already sits in `Aoes` at a measured 87.5%, and in a raid tier near-lethal damage is the design rather than a signal — a threshold that makes sense for dungeon content would put almost everything in this fight in the heavy chain.

**Tested:** Not tested in game. Ids, damage and the marker correlation all come from a live M4 pull, but the fight has not been re-run since the change, so no detection has been observed firing. A wrong id is silent rather than loud here.

**Sources:** Cast ids, head markers, and per-ability damage as a share of target max HP from the combat log of an M4 pull, with the marker-to-mechanic correlation cross-checked against a second fight (Thornmarch) in the same session. Mechanic descriptions cross-checked against the Icy Veins guide, which independently describes Venomous Scourge and Dramatic Lysis as marked AoEs on all players and Fourth-wall Fusion as a stack marker on a healer.

**Removals:** None.

### DIFF
--- a/Magitek/Utilities/FightLogicEncounters.cs
+++ b/Magitek/Utilities/FightLogicEncounters.cs
@@ -7675,12 +7675,9 @@ namespace Magitek.Utilities
                             47552, // Splattershed
                             48096, // Splattershed
                             46208, // Venomous Scourge — marked AoE on every player. Measured 22 hits
-                                   // across all eight at up to 62.2% of a health bar.
-                            46209, // Grotesquerie — inflicts Burning Grotesquerie on all players,
-                                   // which resolves as a marked AoE when it expires. The damage lands
-                                   // ~18s later under a different name (Dramatic Lysis, measured 73.6%
-                                   // on all eight) and has no cast line of its own, so this cast is the
-                                   // only thing that can be reacted to.
+                                   // across all eight at up to 62.2% of a health bar, and its damage
+                                   // lands 5.0s after the cast, so mitigation applied on the cast is
+                                   // still up when the hit arrives.
                         },
                         AoeLockOns = new List<uint> {
                             93,    // Stack marker -> Fourth-wall Fusion, which measured 317.9% of a
--- a/Magitek/Utilities/FightLogicEncounters.cs
+++ b/Magitek/Utilities/FightLogicEncounters.cs
@@ -7674,8 +7674,23 @@ namespace Magitek.Utilities
                             46228, // The Fixer
                             47552, // Splattershed
                             48096, // Splattershed
+                            46208, // Venomous Scourge — marked AoE on every player. Measured 22 hits
+                                   // across all eight at up to 62.2% of a health bar.
+                            46209, // Grotesquerie — inflicts Burning Grotesquerie on all players,
+                                   // which resolves as a marked AoE when it expires. The damage lands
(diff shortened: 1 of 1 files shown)


HomoYouDidnt:
Correct, and the measurements make it worse than described. I have removed Grotesquerie from the entry.

The gap between the Grotesquerie cast and the Dramatic Lysis damage, measured across the pull this was built from:

```
gap 18.2s   gap 32.2s   gap 18.2s   gap 32.2s   gap 18.2s   gap 32.2s   gap 18.2s
```

Against effect durations of 15s for Sacred Soil and 30s for Galvanize, that means the short mitigation has always expired by the time the damage lands, and at the 32.2s variant even the barrier has. On top of that, `DoAndBuffer` adds the cast to `FlHandledCastingSpellId`, so the mechanic is booked as answered and nothing fires again nearer the actual hit — mitigation spent, mechanic marked handled, damage uncovered. Strictly worse than not cataloguing it.

I checked the other addition against the same standard rather than assuming it was fine. Venomous Scourge resolves 5.0s after its cast (five bursts, all 4.9-5.0s), which is inside both effect durations, so mitigation applied at the cast is still up when the hit arrives. That one stays, and I have noted the measured delay in the comment so the reasoning is visible next time.

Your suggestion of detecting the resolution marker or aura timing instead is the right general answer for delayed mechanics, and Fourth-wall Fusion in this same PR is already handled that way via head marker 93 precisely because it has no usable cast. Doing the same for Dramatic Lysis needs the resolution marker identified first, which I do not have from this pull — the markers observed were 654 and 139 in the relevant window, neither confirmed against a second fight, so I would rather leave the mechanic uncovered than guess at its marker.

PR #248, opened 2026-08-19
FightLogic - Mitigate Barnabas and more of Anima's kit in the Tower of Babil

HomoYouDidnt:
**Tested:** SCH Lv100, full Tower of Babil clear 2026-08-19 — but **the entries themselves are not tested in game.** They were derived from that run's ACT log after the fact, so every id below is measured, and none has been seen to *fire* yet. A wrong id fails silently, so absence of a detection line on the next run means the id is wrong rather than the mechanic not happening.

**What prompted it:** Barnabas took the party for 29.4% of a health bar twice with no response. He has no entry at all — the zone lists only Lugae and Anima, one AoE each.

**Detection itself is fine.** Positive control from the same run: both existing ids (`25338` Thermal Suppression, `25344` Mega Graviton) were detected and drew responses. The gaps are catalogue-only.

### Barnabas (10279) — new

- `25324` **Shocking Force** — two casts, each hit all four party members, max 29.4% of a health bar, has its own cast bar.

### Anima (10285 / 10288) — additions

- `25352` + `25351` **Erupting Pain** — 41.1%, all four players, 5.0s from cast to damage.
- `25347` **Boundless Pain** — casts as `25347`, lands as `25348` for 30.0% on all four about 9.2s later, then a long `25349` tail at 5.5%.
- `AoeLockOns = { 139 }` — **this is the part that actually fixes Erupting Pain.** It marks every player 5.0s ahead of the hit. 139 is already on the global `CommonAoeLockOns` set, but `PresentLockOnCounts` only unions that set in when `FightLogicIncludeCommonAoeLockOnsTest` is enabled, so by default relevance comes solely from each enemy's own `AoeLockOns` — and Anima declared none. The marker was seen and discarded.

### Deliberately left out

- **Oblivion `25359` — commented, with a note in the file.** Cast bar ~6.0s, then `23697` ticks 15× across all four at ~10.6% each over 4.3s, and only then `23872` for 49.1% — 13.1s after the cast starts. Reacting on the cast puts a barrier up in time for the chip damage, which eats it before the hit that matters. It needs a per-mechanic reaction delay to be worth cataloguing; the note sketches where that would go (`HodlCastTimeRemaining` already resolves the enemy and every job reaches it) but I have not built it — the shape is your call.
- **Ground and Pound `25322`** (Barnabas, 28.2%, single target, 4 casts) and **Surface Missile `25335`** (Lugae, 20.2%, single target, 8 casts). Both look like tank busters, but three of Ground and Pound's four resolutions carry an empty target field, so I would rather see one clean pull than guess a category.
- Everything under ~2% of a health bar, and the `23697`/`25348`/`25349` damage halves, which have no cast line of their own.

**Sources:** ACT network log 2026-08-19, `Network_30208_20260819.log`, run window 12:51–13:08 local. NpcIds read from AddCombatant; cast bars from type 20; victim counts and damage from types 21/22; the head marker from type 27 field [6] (`008B` = 139).

### DIFF
--- a/Magitek/Utilities/FightLogicEncounters.cs
+++ b/Magitek/Utilities/FightLogicEncounters.cs
@@ -4534,11 +4534,34 @@ namespace Magitek.Utilities
                     }
                 }
             },
+            // Anima's Oblivion (cast id 25359) is left out on purpose. Measured 2026-08-19:
+            // the cast bar runs ~6.0s, then 23697 ticks 15 times across all four players at
+            // ~10.6% of a health bar each over the next 4.3s, and only then does 23872 land for
+            // 49.1% - 13.1s after the cast starts. Reacting on the cast puts a barrier up in time
+            // for the chip damage, which consumes it, so it is gone before the hit that matters;
+            // Sacred Soil's 15s survives but with under 2s to spare.
+            //
+            // What would make it catalogueable is a per-mechanic reaction delay. HodlCastTimeRemaining
+            // already resolves (encounter, enemyLogic, enemy) and every job reaches it - the four
+            // healer HealFightLogic files and CommonFightLogic alike - so a hold keyed on
+            // enemy.CastingSpellId would let this one entry react at cast-end instead of cast-start
+            // without touching any other reaction. Flagging rather than building it: the shape of
+            // that (data on the mechanic vs a lookup in FightLogic) is your call.
             new Encounter {
                 ZoneId = ZoneId.TheTowerOfBabil,
                 Name = "Dungeon: The Tower of Babil",
                 Expansion = FfxivExpansion.Endwalker,
                 Enemies = new List<Enemy> {
+                    new Enemy {
+                        Id = 10279,
+                        Name = "Barnabas",
+                        TankBusters = null,
+                        SharedTankBusters = null,
+                        Aoes = new List<uint> {
+                            25324 //Shocking Force
+                        },
+                        BigAoes = null
+                    },
                     new Enemy {
                         Id = 10281,
                         Name = "Lugae",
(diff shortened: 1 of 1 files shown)


HomoYouDidnt:
Follow-up from checking the two I had left unclassified against XIVAPI v2 targeting data — it corrected me on both, and turned up something I had missed in the entry itself.

**Barnabas needed a lock-on too.** `f6687579` adds `AoeLockOns = { 62 }`. Shocking Force is a **stack**, not a plain party AoE:

```
12:52:46.226  marker 62 on Vasper Sunday   (one player marked)
12:52:46      cast 25324 targeted at that player
12:52:51      4 resolutions - all four of us
```

Same shape again at 12:53:30. Marker 62 is the stack marker and it is on the global `CommonAoeLockOns` set — which, as with Anima and 139, is inert unless the enemy declares it. Without this the routine sees the marker 5.0s ahead of the hit and discards it.

**Ground and Pound `25322` — my "looks like a tank buster" was wrong.** `CastType=12, EffectRange=40, Range=0, CanTargetSelf=true`, and the log agrees: the cast targets Barnabas himself, three of its four resolutions hit nobody at all, and the one that landed hit the tank for 28.2%. It is a proximity AoE that catches whoever is in melee, not a buster — a ranged healer takes zero from it. Still excluded, but as a positional rather than as an unclassified buster.

**Surface Missile `25335`** — `CastType=2, EffectRange=6, Range=100`. A 6-yalm ground circle; the single victim was one person standing in it. A dodgeable puddle, correctly left out.

Worth recording for its own sake: the ACT victim counts were misleading on both, and the targeting fields settled them in about a minute. A self-targeted cast logs resolutions with an EMPTY target field, which reads as "one victim" if you only count distinct targets.

PR #250, opened 2026-08-19
FightLogic - Mitigate the stacks and marked AoEs in Anamnesis Anyder

HomoYouDidnt:
**Tested:** SCH Lv100, full Anamnesis Anyder clear 2026-08-19. The entries themselves are **not tested in game** — measured from that run's ACT log afterwards, so every id is real but none has been seen to fire yet.

**Why:** the zone had one AoE for Kyklops and nothing but a tank buster for Rukshs Dheem. Seven casts of a party-wide hit went unanswered in a single clear.

### Unknown `9260` / `9261`

- `19322` **Ectoplasmic Ray** — a stack. Damage lands as `19320` on all four players 5.25s after the cast, up to 30.4%.

### Kyklops `9263`

- `19296` **Open Hearth** — stack marker on one player, 4 hits from one cast on all four, up to 33.1%.
- `AoeLockOns = { 62, 96 }` — Open Hearth's stack marker, and Wanderer's Pyre's spread marker (`96`, placed on three players in the same millisecond as its cast). Both are on the global `CommonAoeLockOns` set, which is inert unless the enemy declares it, so neither did anything before.

### Rukshs Dheem `9264`

- `19323` **Seabed Ceremony** — the big one. **Seven casts in one clear**, every one hitting all four players, up to 41.7%. Damage id `19324`; `19323` is the self-targeted telegraph, cast in the same millisecond, ~4.0s lead.
- `19325` **Falling Water** — marks two random players at once (damage `19326`), and not the tanks. Filed under `Aoes` rather than `SharedTankBusters` deliberately: the targets are random, so the party answer is the right one.
- `19327` **Flying Fount** — a stack (damage `19328`), up to 52.1%.
- `AoeLockOns = { 62 }` — Flying Fount's stack marker.

### Deliberately excluded

Nearly everything else in this dungeon is a dodge, and the hits-per-cast ratio separates them cleanly:

| | hits/cast |
|---|---|
| Open Hearth, the Final Verse | 4.0, 3.7 |
| 2,000-mina swing, 2,000-mina swipe | 1.5, 1.0 |
| Raging Glower, Eye of the Cyclone, Terrible Hammer, Terrible Blade | 0.0 |

Swing reads 72.9% and swipe 36.9%, which is why damage magnitude is the wrong sort key — both are avoidable, and the duty guide agrees ("watch the boss's cast bar to avoid"). Also excluded: Scrutiny, Luminous Ray, Clearout, Setback, Command Current, Rising Tide and both Depth Grip variants, all positional.

**Note on markers:** Ectoplasmic Ray is a stack but ACT logged **no** head marker for it — three type-27 lines exist in the whole Unknown fight and all three are the Fetid Fang telegraph. Open Hearth and Flying Fount do log marker 62. The channel is inconsistent, so every entry here reacts to the cast id, with `AoeLockOns` as a bonus where a marker was actually observed.

**Sources:** ACT network log `Network_30208_20260819.log`, run window 16:34–16:56 local. NpcIds from AddCombatant field [9]; cast bars from type 20; damage and victim counts from types 21/22 filtered to damage-flagged lines landing on players; markers from type 27 field [6]. Classifications cross-checked against the duty guide and confirmed by the player mid-run.

### DIFF
--- a/Magitek/Utilities/FightLogicEncounters.cs
+++ b/Magitek/Utilities/FightLogicEncounters.cs
@@ -2483,7 +2483,9 @@ namespace Magitek.Utilities
                         },
                         SharedTankBusters = null,
                         Aoes = new List<uint>() {
-                            19306 //Inscrutability
+                            19306, //Inscrutability
+                            19322 //Ectoplasmic Ray - stack. Damage lands as 19320 on all four
+                                  //players 5.25s after this cast, up to 30.4% of a health bar.
                         },
                         BigAoes = null
                     },
@@ -2496,7 +2498,9 @@ namespace Magitek.Utilities
                         },
                         SharedTankBusters = null,
                         Aoes = new List<uint>() {
-                            19306 //Inscrutability
+                            19306, //Inscrutability
+                            19322 //Ectoplasmic Ray - stack. Damage lands as 19320 on all four
+                                  //players 5.25s after this cast, up to 30.4% of a health bar.
                         },
                         BigAoes = null
                     },
@@ -2506,7 +2510,13 @@ namespace Magitek.Utilities
                         TankBusters = null,
                         SharedTankBusters = null,
                         Aoes = new List<uint>() {
-                            19288 //The Final Verse
+                            19288, //The Final Verse
+                            19296 //Open Hearth - 4 hits from one cast, all four players, 33.1%
+                        },
+                        AoeLockOns = new List<uint>() {
+                            96 //Wanderer's Pyre marks three players in the same millisecond as
+                               //its cast. On the global common-marker set, which is inert unless
+                               //the enemy declares it.
                         },
                         BigAoes = null
                     },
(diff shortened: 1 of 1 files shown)


HomoYouDidnt:
Pushed **cde5a27e** after a second Anamnesis Anyder run on 2026-08-20.

### Each of these casts is a pair, and I had only listed half

Seabed Ceremony, Falling Water and Flying Fount each fire two ids in the same window, sharing one cast time: a self-targeted id with `EffectRange 0`, and the id that carries the geometry and the damage.

| mechanic | r0 half | effect half |
|---|---|---|
| Seabed Ceremony | `19323` single r0, 4.0s — 7 casts, **0 damage** | `19324` circle r60, 4.0s — 7 casts, 7 hits, 41.3% |
| Falling Water | `19325` single r0, 5.0s — 3 casts, **0 damage** | `19326` circle r8, 5.0s — 6 casts |
| Flying Fount | `19327` single r0, 5.0s — 1 cast, **0 damage** | `19328` circle r6, 5.0s — 1 cast, 1 hit, 38.1% |

The original entry listed only the r0 halves and mentioned the effect ids in comments. Both halves show a cast bar, and I cannot tell from a log which one the client reports as the enemy's current cast, so both are now listed — the same way the Clyteum entry handles this pattern. A non-matching id is inert rather than harmful.

Shapes are from XIVAPI `CastType`/`EffectRange`/`Cast100ms`; counts from the ACT log for the 09:47–10:37 local run.

### Io Ousia — one trash mob, added on a player ruling

`20004` Barreling Smash is a lock-on charge along a line at whoever it picks, measured at 30.5% on its single victim. It is filed under `TankBusters` because that is the path that shields the *named target* — the victim is not necessarily a tank. No head marker precedes it (checked every type-27 line in the 8s before each cast), so the 5.0s cast bar is the only signal available.

This is the only non-boss entry in the encounter, which is deliberate rather than an oversight.

**Tested:** SCH Lv100, two full clears. The ids themselves are measured but have not been seen to fire — they are not in my test build.

PR #260, opened 2026-08-20
RDM - Fix Magicked Swordplay logic to allow melee combo initiation

lokibunny:
### Summary of Changes
* **Magicked Swordplay Fix:** Corrected state-checking logic to ensure the routine properly recognizes and consumes stacks to initiate the melee combo sequence and combo finishers.

### DIFF
--- a/Magitek/Logic/RedMage/SingleTarget.cs
+++ b/Magitek/Logic/RedMage/SingleTarget.cs
@@ -42,7 +42,23 @@ namespace Magitek.Logic.RedMage
             if (InAoeCombo() || (AoeControl.Enabled && Core.Me.EnemiesInCone(8) >= RedMageSettings.Instance.AoeEnemies))
                 return false;
 
-            if (!Core.Me.HasAura(Auras.MagickedSwordplay))
+            // If Magicked Swordplay is active, ensure we actually want to spend it, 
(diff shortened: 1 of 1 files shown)


cheesegoldfish:
Closing this. It stops Riposte from casting at all.

SingleTarget.cs:55 needs ActionManager.LastSpell to be Zwerchhau to pass.
SingleTarget.cs:58 needs it to be Riposte. LastSpell is one value and both have to
pass, so Riposte returns false every time Magicked Swordplay is up. Magicked
Swordplay only comes from Manafication at 60, so Zwerchhau and Redoublement are
always known and both guards are always armed.

Ran your branch on a dummy. Aura log:

    -75.4s  gained  Magicked Swordplay  stacks 3  timeLeft 30.0
    -45.0s  lost    Magicked Swordplay  stacks 3  timeLeft  0.2

Full 30s, all 3 stacks expired unused, combo never started. Embolden went out 1.1s
before Manafication so its cooldown sat between 119s and 88s the whole window,
nowhere near HoldAccelForEmboldenSeconds. Nothing else was blocking it.

The rest of the diff does nothing. The else branch is line-for-line what was already
there. The Embolden check at :50 is a copy of the one at :37 that already returned.

The melee combo works on master.

PR #267, opened 2026-08-23
WHM - Balance optimizations and logic fixes

lokibunny:
Description
This PR focuses strictly on White Mage logic improvements, particularly around Dawntrail skill interactions, cooldown weaving, and proactive shield management.

Key Changes
Glare IV & Presence of Mind (Buff.cs, SingleTarget.cs):

Added Sacred Sight aura and target safety checks for Glare IV.

Fixed Presence of Mind conditions to account for Glare IV being a GCD, allowing PoM to properly weave while the GCD is rolling rather than hanging.

Divine Caress Optimization (Heal.cs):

Implemented holding logic for Divine Caress. The routine will now hold the ability if Temperance is active to better chain mitigations and shields, only casting it automatically if the Divine Grace buff is within 5 seconds of expiring.

General Weaving & Cooldowns:

Removed the obsolete Dia restriction on Divine Benison to allow free weaving with the 1.5s Glare cast time.

Removed the DoT duration check on Assize to prevent unnecessarily delaying damage.

### DIFF
--- a/Magitek/Logic/WhiteMage/Buff.cs
+++ b/Magitek/Logic/WhiteMage/Buff.cs
@@ -43,9 +43,12 @@ namespace Magitek.Logic.WhiteMage
             if (!Core.Me.InCombat)
                 return false;
 
+            // Reason: Glare IV is a GCD, so its cooldown is the GCD timer. 
+            // Presence of Mind is an oGCD we want to weave while the GCD is rolling, 
+            // which is exactly when GlareIV.IsKnownAndReady() evaluates to false.
             if (WhiteMageSettings.Instance.PresenceOfMindForGlareIV
                 && Spells.PresenceofMind.IsKnownAndReady()
-                && Spells.GlareIV.IsKnownAndReady())
+                && Spells.GlareIV.IsKnown())
             {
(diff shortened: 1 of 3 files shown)


cheesegoldfish:
Taking this over as #281 so the WHM fixes that are wrong don't ship. Your commit is in that branch, so you're credited on the release.

Kept: the Divine Caress hold, now behind a setting that defaults to off. Divine Caress is also our fightlogic shield response — when a raidwide is inbound we want the shield on the party, not held for spacing. Off by default, on for anyone who wants the chain.

Dropped, Presence of Mind: Glare IV is a GCD, so IsKnownAndReady is doing real work there and IsKnown() removes it. Verified in game.

Dropped, Glare IV target check: Rotations/WhiteMage.cs:169 already gates every damage path on ThoroughCanAttack, which is a null check plus CanAttack plus CanBeDamagedByMe. Also, re-reading Core.Me.CurrentTarget three times doesn't survive a mid-pulse death — see ee1045cf, which fixed exactly that in this file by capturing the target once.

Two of the four items in your description aren't in the diff — the Divine Benison Dia restriction and the Assize DoT check are both still on master, untouched. Worth checking what you pushed before opening.

PR #269, opened 2026-08-23
SCH - Optimize Aetherflow usage and fairy gauge logic

lokibunny:
Description
This PR isolates logic improvements specifically for Scholar, focusing on better resource management for both Aetherflow stacks and the Fairy Gauge to prevent overcapping and improve overall healing/damage efficiency.

Key Changes
Aetherflow Optimization: Refined the logic for spending Aetherflow stacks (e.g., Energy Drain) to ensure they are fully utilized before Aetherflow comes off cooldown, preventing lost stacks and overcapping.

Fairy Gauge Management: Improved the conditions for triggering Fairy Gauge abilities, ensuring the gauge is spent efficiently during combat rather than sitting at maximum capacity unused.


cheesegoldfish:
Something is wrong with this PR, it's trying to undo all of the optimizations that were recently put in a couple of days ago.

PR #276, opened 2026-08-24
FightLogic - Respond to a tankbuster on whoever it actually targets, not only tanks

HomoYouDidnt:
## What this changes

The tankbuster matcher now returns the castable ally the buster is actually aimed at, instead of only matching when that ally is a tank.

## Why

Two real cases went silent under the tank-only filter: a dead main tank retargets the buster onto whoever has aggro next, and some catalogued busters (Io Ousia's Barreling Smash) pick an arbitrary player by design. In both, the response was needed *more* than usual and fired not at all.

**Anything removed / widened:** this deliberately widens a filter - buster responses can now fire for a non-tank victim. It still only matches catalogued buster ids aimed at a castable ally, and the answer-once suppression is unchanged, so the widening adds responses only in exactly the cases that were silent.

**Tested:** running in daily play on my test build since 2026-08-22 across raids, dungeons, and field content; responses observed following the actual victim, including non-tank cases.

**Sources:** the Barreling Smash arbitrary-target behavior observed in combat logs (the catalogued id landing on non-tanks with the main tank alive).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Magitek/Logic/Roles/CommonFightLogic.cs
+++ b/Magitek/Logic/Roles/CommonFightLogic.cs
@@ -18,7 +18,11 @@ namespace Magitek.Logic.Roles
             if (!FightLogic.ZoneHasFightLogic() || !FightLogic.EnemyHasAnyTankbusterLogic())
                 return false;
 
-            if (FightLogic.EnemyIsCastingTankBuster() != null
+            // Personal defensives only answer a hit that will land on US: a single-target
+            // buster aimed at someone else (now matchable since the victim widening) must not
+            // burn our cooldowns. A shared buster hits every tank, so it always qualifies.
+            var busterVictim = FightLogic.EnemyIsCastingTankBuster();
+            if ((busterVictim != null && busterVictim == Core.Me)
                 || FightLogic.EnemyIsCastingSharedTankBuster() != null)
             {
                 if (Core.Me.HasAnyAura(defensiveAuras))
--- a/Magitek/Utilities/FightLogic.cs
+++ b/Magitek/Utilities/FightLogic.cs
@@ -282,8 +282,12 @@ namespace Magitek.Utilities
 
         private static Character MatchTankBuster(Enemy enemyLogic, BattleCharacter enemy)
         {
+            // The victim is whoever the cast is aimed at - usually a tank, but not always: a dead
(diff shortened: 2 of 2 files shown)


cheesegoldfish:
Keep the `== Core.Me` gate, that one's right. Every spell in the four tank Defensives arrays is self-cast.

The range one you declined on a wrong fact. `CastableTanks` is not built from the ≤30y list. Group.cs:237 adds to `CastableTanks`, Group.cs:243 is where `distance` first gets computed — six lines later. It has no distance filter. `MatchSharedTankBuster`, directly below the function you edited, says so in its own comment.

Before: unbounded. After: raw ≤30y. That's a loss, not pre-existing bounds.

Codex also wasn't asking for a Group change. It asked for `WithinSpellRange(30)` on `CastableParty` inside the matcher, same as the shared-buster matcher does. Three lines.

Make it two-tier:

```csharp
Group.CastableTanks.FirstOrDefault(x => x == enemy.TargetCharacter)
  ?? Group.CastableParty.FirstOrDefault(x => x == enemy.TargetCharacter && x.WithinSpellRange(30))
```

Tier 1 is master verbatim, so tanks are unchanged and the reachability argument goes away. Tier 2 only fires where master returned null, so it can't regress the other call sites. Still fixes the dead-tank case you opened this for.

`MatchTankBuster` isn't just TankDefensive anymore. `FightLogic_Debuff` puts Feint/Addle/Reprisal on the caster, not the victim, so a victim-range filter there only loses responses.

Apply that and I'll merge. Resolving the Codex threads.

HomoYouDidnt:
Applied exactly as prescribed in a615d0b6 — tier 1 is master's CastableTanks lookup verbatim, tier 2 adds CastableParty with WithinSpellRange(30) only where master returned null. Built clean.

And you're right about the decline: I asserted CastableTanks was distance-bounded without reading Group.cs, when the answer was sitting six lines above the distance computation — and in MatchSharedTankBuster's own comment directly below the function I was editing. Codex's finding was correct and my refutation was the wrong-fact one. That's exactly the failure mode I'm supposed to be checking for before replying to review, and it's been recorded on my end so the next decline comes with the source actually read.

PR #288, opened 2026-08-26
AST - Lord of Crowns only fires in combat, at enemies actually around you

HomoYouDidnt:
Small follow-up from live observation: Lord of Crowns fired 0.7s after out-of-combat raises, five times across two days, once onto an idle pack.

**Mechanism:** the heal-oGCD block is gated on `InActiveDuty || InCombat`, and `InActiveDuty` stays true for the whole instance — so between pulls the block still runs, and Lord's guards had no combat check. Its enemy-count guard also skipped counting entirely at the default setting (`> 1` short-circuit) and counted around `CurrentTarget` (which often does not exist between pulls, and is the wrong centre anyway — Lord is self-centred).

**Fix:** an in-combat gate, and the count is now around the caster with a floor of one enemy in range regardless of the setting.

**Tested:** builds clean; the firing pattern was measured from ACT logs (AST Lv100, duty + Occult Crescent). The fix itself is **not yet exercised in game** — the census check after the next AST session is 'zero Lord casts outside combat'.

**Removals:** the `LordOfCrownsEnemies > 1` short-circuit and the target-centred count, both replaced as described.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cheesegoldfish (changes requested):
Three follow-ups, all in `LordOfCrown()`. The in-combat gate is right and matches `Macrocosmos` / `EarthlyStar` / `CollectiveUnconscious`, which already carry the same guard — Lord was the only member of that block without one. The `EnemiesNearby` swap is correct and matches AGENTS.md.

What's left is what the old guard was hiding.

cheesegoldfish (approved):

